PR #7, opened 2014-07-14
PhpAdapter supports prevention of array merging

vojtech-dobes:
Allows to use `!` trick in array's key to force overriding instead of merging.


TomasVotruba:
What needs to be done here?

PR #22, opened 2014-10-06
ContainerBuilder::findByType() returns all services … [WIP]

### LINKED ISSUE #21: Better API for ContainerBuilder::findByType
`ContainerBuilder::findByType` and `Container::findByType` works similar:

- returns names of services (because Container cannot return objects)
- by default they return only autowired services

On the one hand it is good that these two functions behave consistently, but on the other hand, it is uncomfortable to use ContainerBuilder::findByType:

- you can always need retrieve definition using `$builder->getDefinition($name)`
- it's not clear what FALSE means (classic boolean arguments problem)

So I think that `ContainerBuilder::findByType` should return all services as pairs `[name => definition]`. 

The question is whether `Container::findByType` can be in future changed to return names of all (including non-autowired) services too? Is it big BC?

(`ContainerBuilder::findByType` and second argument of `Container::findByType` exist only in dev-version).

dg:
Closes #21

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -179,18 +179,23 @@ class ContainerBuilder extends Nette\Object
 
 
 	/**
-	 * Gets the service names of the specified type.
+	 * Gets the service names and definitions of the specified type.
 	 * @param string
-	 * @return string[]
+	 * @return ServiceDefinition[]
 	 */
-	public function findByType($class, $autowired = TRUE)
+	public function findByType($class)
 	{
 		$class = ltrim($class, '\\');
 		self::checkCase($class);
-		return array_merge(
-			isset($this->classes[$class][TRUE]) ? $this->classes[$class][TRUE] : array(),
-			!$autowired && isset($this->classes[$class][FALSE]) ? $this->classes[$class][FALSE] : array()
-		);
+		$found = array();
+		foreach (array(TRUE, FALSE) as $mode) {
+			if (!empty($this->classes[$class][$mode])) {
+				foreach ($this->classes[$class][$mode] as $name) {
+					$found[$name] = $this->definitions[$name];
+				}
+			}
+		}
+		return $found;
 	}
 
 
--- a/src/Bridges/DITracy/templates/ContainerPanel.panel.phtml
+++ b/src/Bridges/DITracy/templates/ContainerPanel.panel.phtml
@@ -45,7 +45,7 @@ use Nette,
 		</thead>
 		<tbody>
 		<?php foreach ($services as $name => $class): ?>
-		<?php $autowired = in_array($name, $container->findByType($class)); ?>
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
 		}
 	],
 	"require": {
-		"php": ">=5.3.1",
+		"php": ">=5.3.7",
```
dg: Nette will keep 5.3.3 compatibility till January 2016 in 2.2 branch.

dg:
This is the last thing that must be considered before nette/di can become RC. 

//cc @fprochazka @JanTvrdik @VasekPurchart @milo @Majkl578 @Vrtak-CZ …

(Note that ContainerBuilder::findByType and second argument of Container::findByType exist only in dev-version).

fprochazka:
I'm using it only in [PresentersLocator](https://github.com/Kdyby/PresentersLocator) which I'm going to deprecate anyway, as it's going to be useless with Nette 2.3. The `Container::findByType` shouldn't also be a problem for me. 

Now if I wanna get one autowired service of given type, I'll use `getByType()` and if I need to search for the type in all service, I'll be using `findByType()`, right?

src/DI/ContainerBuilder.php
```diff
@@ -178,17 +178,22 @@ public function getByType($class)
...
-		self::checkCase($class);
-		return array_merge(
-			isset($this->classes[$class][TRUE]) ? $this->classes[$class][TRUE] : array(),
-			!$autowired && isset($this->classes[$class][FALSE]) ? $this->classes[$class][FALSE] : array()
-		);
+		$is_subclass_of = PHP_VERSION_ID >= 50307 ? 'is_subclass_of' : function($class, $of) {
+			return array_key_exists($of, class_parents($class) + class_implements($class));
+		};
+		$found = array();
+		foreach ($this->definitions as $name => $def) {
```
fprochazka: I'm a bit concerned about the performance of this method. Also, it still depends on the `prepareClassList`, because it can't resolve types from `@return` annotations by itself.

dg: Both implementations (`foreach & $this->definitions` or `isset($this->classes)`) depends on the prepareClassList. But you are right that this is slower.

dg:
Autowiring is about *THE ONE* instance, so findByType limited to autowired services is IMHO flawed.

fprochazka:
@dg makes sense :)

composer.json
milo: @dg OT: Current Debian stable Wheeze has 5.4.36. New stable Jessie should be out in March with 5.6.5.

milo:
Hard to say an opinion since I never used the `findByType()`. Maybe when someone needed all autowired services of given interface/parent? (Don't know real world use-case.)

VasekPurchart:
+1

PR #29, opened 2014-11-17
ContainerBuilder: Add support for aliasing service definitions

xificurk:
At the moment there is no sensible way how to rename a service without massive BC break, that's why some of the "Nette" services use prefixes, some not etc. This PR tries to fix this by adding support for service aliases to `ContainerBuilder`.
Open question is whether or not to automatically trigger deprecation warning when the old name (=alias) is used instead of the new one.

### DIFF
--- a/src/DI/Container.php
+++ b/src/DI/Container.php
@@ -20,6 +20,7 @@ class Container extends Nette\Object
 	const TAGS = 'tags';
 	const TYPES = 'types';
 	const SERVICES = 'services';
+	const ALIASES = 'aliases';
 
 	/** @var array  user parameters */
 	/*private*/public $parameters = array();
@@ -60,7 +61,9 @@ class Container extends Nette\Object
 		if (!is_string($name) || !$name) {
 			throw new Nette\InvalidArgumentException(sprintf('Service name must be a non-empty string, %s given.', gettype($name)));
 
-		} elseif (isset($this->registry[$name])) {
+		}
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
+		if (isset($this->registry[$name])) {
 			throw new Nette\InvalidStateException("Service '$name' already exists.");
 
 		} elseif (!is_object($service)) {
@@ -82,6 +85,7 @@ class Container extends Nette\Object
 	 */
 	public function removeService($name)
 	{
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		unset($this->registry[$name]);
 	}
 
@@ -94,6 +98,7 @@ class Container extends Nette\Object
 	 */
 	public function getService($name)
 	{
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		if (!isset($this->registry[$name])) {
 			$this->registry[$name] = $this->createService($name);
 		}
@@ -108,6 +113,7 @@ class Container extends Nette\Object
 	 */
 	public function hasService($name)
 	{
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		return isset($this->registry[$name])
 			|| method_exists($this, $method = Container::getMethodName($name)) && $this->getReflection()->getMethod($method)->getName() === $method;
 	}
@@ -123,6 +129,7 @@ class Container extends Nette\Object
 		if (!$this->hasService($name)) {
 			throw new MissingServiceException("Service '$name' not found.");
 		}
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		return isset($this->registry[$name]);
 	}
 
@@ -135,6 +142,7 @@ class Container extends Nette\Object
 	 */
 	public function createService($name, array $args = array())
 	{
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		$method = Container::getMethodName($name);
 		if (isset($this->creating[$name])) {
 			throw new Nette\InvalidStateException(sprintf('Circular reference detected for services: %s.', implode(', ', array_keys($this->creating))));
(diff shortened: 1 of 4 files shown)


enumag:
:+1:

dg:
Direct calling of `$container->createService()` is not desired action, so I think that we may solve it via magic __call instead of creating new methods.

src/DI/Container.php
```diff
@@ -123,6 +129,7 @@ public function isCreated($name)
 		if (!$this->hasService($name)) {
 			throw new MissingServiceException("Service '$name' not found.");
 		}
+		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		return isset($this->registry[$name]);
 	}
```
dg: What about Container::createService() ?

src/DI/ContainerBuilder.php
```diff
@@ -49,6 +52,38 @@ class ContainerBuilder extends Nette\Object
 
 
 	/**
+	 * @param string
+	 * @param string
+	 */
+	public function addAlias($alias, $service)
```
dg: Can you move it after hasDefinition()?

src/DI/ContainerBuilder.php
```diff
@@ -755,6 +814,7 @@ public function getServiceName($arg)
 			}
 			return $res;
 		}
+		$service = isset($this->aliases[$service]) ? $this->aliases[$service] : $service;
 		if (!isset($this->definitions[$service])) {
```
dg: !hasDefinition($service) seems better here

src/DI/ContainerBuilder.php
xificurk: done

src/DI/ContainerBuilder.php
xificurk: Imho `getServiceName()` should resolve aliases and return "canonical" service name. Call to `!hasDefinition($service)` is then redundant.

src/DI/ContainerBuilder.php
```diff
@@ -57,8 +60,9 @@ public function addDefinition($name, ServiceDefinition $definition = NULL)
 	{
 		if (!is_string($name) || !$name) { // builder is not ready for falsy names such as '0'
 			throw new Nette\InvalidArgumentException(sprintf('Service name must be a non-empty string, %s given.', gettype($name)));
-
-		} elseif (isset($this->definitions[$name])) {
+		}
+		$name = isset($this->aliases[$name]) ? $this->aliases[$name] : $name;
```
dg: Maybe addDefinition() should be supposed to create only "original" services, so what about to throw exception when `if (isset($this->definitions[$name]) || isset($this->aliases[$name]))`?

src/DI/ContainerBuilder.php
dg: You're right.

src/DI/Container.php
xificurk: Fixed

xificurk:
@dg As I started implementing your suggestion about magic `__call()` in `Container` I realized that it's not so simple. The mapping between service name and method name is not 1:1, e.g. `'my.service'` and `'my__service'` are both mapped to `createServiceMy__service()`. Although this is an edge case, I think it's better to stay on the safe side... plus, it's less magic.

src/DI/ContainerBuilder.php
xificurk: I think that would break `Compiler::parseServices()` logic.

dg:
I know, double `__` is forbidden in service name (but it is not checked), so you can rely on that. 

On other side, the best solution in to not support it at all. In NetteExtension there is no service which is has sense to create using createServiceMethod. But above all, it is not supported practice, so I am for to not implemented it, at least not in this commit.

xificurk:
Alright, which solution do you prefer? With `__call` fallback, or completely remove support for `createServiceAlias()`? I guess I'm fine with both of those.

dg:
Both, `createService('...')` is enough.

xificurk:
@dg I meant - should I implement magic `__call()` method in `Container` as fallback for creating aliased services, or is the current code ok?

What's your take on whether or not to automatically trigger deprecation warning when the old name (=alias) is used instead of the new one?

src/DI/ContainerBuilder.php
```diff
@@ -101,16 +107,49 @@ public function getDefinitions()
...
+		if (!is_string($alias) || !$alias) { // builder is not ready for falsy names such as '0'
+			throw new Nette\InvalidArgumentException(sprintf('Alias name must be a non-empty string, %s given.', gettype($alias)));
+
+		} elseif (isset($this->aliases[$alias])) {
+			throw new Nette\InvalidStateException("Alias '$alias' has already been added.");
+
+		} elseif (isset($this->definitions[$alias])) {
+			throw new Nette\InvalidStateException("Service '$alias' has already been added.");
+
+		} elseif (!isset($this->definitions[$service])) {
```
dg: This check is unnecessary, because you can add definition, add alias, remove definition.

src/DI/ContainerBuilder.php
dg: Good point

dg:
I am for no deprecation warnings at least in 2.3.

src/DI/ContainerBuilder.php
```diff
@@ -480,6 +520,18 @@ public function generateClasses($className = 'Container', $parentName = NULL)
 			}
 		}
 
+		foreach ($aliases as $alias => $service) {
```
dg: Why not simply `$meta->value[Container::ALIASES] = $aliases`?

xificurk: Hm... originally it was to exclude aliases for non-defined services, but now I realize that we might actually want to keep those. And omitting the check for sane value of alias is probably no big deal.

dg:
Great work! Can you squash it, I'll merge it.

xificurk:
@dg Great, do you want it all in one big commit?

dg:
Yes, this is atomic change. Btw there is one failing test.

xificurk:
Yeah, I noticed that, but it seems unrelated... I'll need to investigate it further.

xificurk:
OK, it's ready now.

PR #33, opened 2015-01-06
Fixed usage of inject annotations in traits

enumag:


enumag:
@dg You merged it too quickly, according to travis code checker complains about the new test filename.

dg:
Hmm, I'll repush it.

PR #34, opened 2015-01-17
ContainerBuilder: fixed return annotation of getDefinitions()

fprochazka:


dg:
Why `array`?

fprochazka:
To be honest, I wasn't looking very well to find existing convention. And in fact it does return an array so I thought it's not a big deal. I just simply added the typehint.

Is it better now?

fprochazka:
Thank you!

PR #35, opened 2015-01-19
Compiler: aliases for nette config section

dg:
This is hardcoded hack for nette/bootstrap#4, which makes every new extension much more readable and usable.


xificurk:
Just a few notes:
- I did not check every part of old `NetteExtension`, but e.g. `nette.mailer` section chagend to `mail`, not `mailer`.
- This requires #29 and the new extensions must use it.

PR #36, opened 2015-01-19
CompilerExtension::validateConfig() returns combined array

dg:
I have a problem with validateConfig() usage, it is not understandable. Typical usecase is:

```php
$config = $this->getConfig($this->defaults);
$this->validateConfig($this->defaults, $config);
```

In this case `$config` is unnecessary, so you can write: 

```php
$config = $this->getConfig($this->defaults);
$this->validateConfig($this->defaults);
```

It seems like WTF. Defaults are passed to `validateConfig` for the same reasons as to `getConfig()`, but it is really weirg. 

What about this simplification?

```php
$config = $this->validateConfig($this->defaults);
```

Or

```php
$config = $this->getValidatedConfig($this->defaults);
```

### DIFF
--- a/src/DI/CompilerExtension.php
+++ b/src/DI/CompilerExtension.php
@@ -46,12 +46,14 @@ abstract class CompilerExtension extends Nette\Object
 
 	/**
 	 * Returns extension configuration.
-	 * @param  array default unexpanded values.
 	 * @return array
 	 */
-	public function getConfig(array $defaults = NULL)
+	public function getConfig()
 	{
-		return Config\Helpers::merge($this->config, $this->getContainerBuilder()->expand($defaults));
+		if (func_num_args()) { // deprecated
+			return Config\Helpers::merge($this->config, $this->getContainerBuilder()->expand(func_get_arg(0)));
+		}
+		return $this->config;
 	}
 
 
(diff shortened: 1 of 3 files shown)


lookyman:
:+1:

dg:
It is not possible due back compatiblity.

fprochazka:
@lookyman I would be strongly against such API. Unless there is a strong and complex validation framework in place, I don't wanna tie my hands with such a low level validation.

lookyman:
@fprochazka Now that I can agree with.

I'd be OK with both method names @dg proposed, I like the first one better.

JanTvrdik:
`getValidatedConfig` is certainly a better name if it returns the validated config

dg:
What about `getTidyConfig()` or `getAdjustedConfig()`?

lookyman:
Well, it doesn't tidy or adjust anything. It just throws an exception if it doesn't recognize a config key.

Btw I was proposing a BC break simply because I feel that with this change, `getConfig()` will eventually just fall out of use, in favor of this new method.

dg:
I am not proposing BC break, validateConfig exists only in dev.

milo:
```
$config = $this->mergeConfig($this->defaults);
$config = $this->getUnifiedConfig($this->defaults);
```

fprochazka:
`mergeConfig` is nice :+1:

dg:
Is expectable that it throws exception, when config contains items which are not in first parametr?

dg:
…`processConfig()`?

PR #37, opened 2015-01-19
ContainerFactory: fixed thread-safety, maybe

JanTvrdik:
A naive port of https://github.com/nette/latte/pull/49 attempting to solve the [described issues](https://github.com/nette/latte/pull/49#issuecomment-70537717).

### DIFF
--- a/src/DI/ContainerFactory.php
+++ b/src/DI/ContainerFactory.php
@@ -110,44 +110,54 @@ class ContainerFactory extends Nette\Object
 	{
 		$key = md5(serialize(array($this->config, $this->configFiles, $this->class, $this->parentClass)));
 		$file = "$this->tempDirectory/$key.php";
-
-		if (!$this->autoRebuild && (@include $file) !== FALSE) { // @ - file may not exist
+		if (!$this->isExpired($file) && (@include $file) !== FALSE) {
 			return;
 		}
 
-		$handle = fopen("$file.lock", 'c+');
-		if (!$handle) {
-			throw new Nette\IOException("Unable to open or create file '$file.lock'.");
+		if (!is_dir($this->tempDirectory)) {
+			@mkdir($this->tempDirectory); // @ - directory may already exist
 		}
 
+		$handle = fopen("$file.lock", 'c+');
+		if (!$handle || !flock($handle, LOCK_EX)) {
+			throw new Nette\IOException("Unable to acquire exclusive lock '$file.lock'.");
+		}
+
+		if (!is_file($file) || $this->isExpired($file)) {
+			$this->dependencies = array();
+			$code = $this->generateCode();
+			$meta = serialize(array_combine($this->dependencies, @array_map('filemtime', $this->dependencies)));
+			if (file_put_contents("$file.tmp", $meta) !== strlen($meta) || !rename("$file.tmp", "$file.meta")) {
+				@unlink("$file.tmp"); // @ - file may not exist
+				throw new Nette\IOException("Unable to create file '$file.meta'.");
+			}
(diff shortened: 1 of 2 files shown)


JanTvrdik:
This seems to be lacking attention.

dg:
Is it ready to merge?

JanTvrdik:
It is very close. I was hoping that somebody could review it and say if it looks fine. I have not given this nearly as much time as I have given to the Latte implementation.

src/DI/ContainerLoader.php
```diff
@@ -61,46 +61,57 @@ public function getClassName($key)
...
+
+			$dependencies = (array) $dependencies;
+			$meta = serialize(array_combine($dependencies, @array_map('filemtime', $dependencies)));
+			if (file_put_contents("$file.tmp", $meta) !== strlen($meta) || !rename("$file.tmp", "$file.meta")) {
+				@unlink("$file.tmp"); // @ - file may not exist
+				throw new Nette\IOException("Unable to create file '$file.meta'.");
+			}
+
+			$code = "<?php\n" . $code;
+			if (file_put_contents("$file.tmp", $code) !== strlen($code) || !rename("$file.tmp", $file)) {
```
dg: What about something like

`if (file_put_contents("$file.tmp1", $meta) !== strlen($meta) || file_put_contents("$file.tmp2", $code) !== strlen($code) 
|| !rename("$file.tmp1", "$file.meta") || !rename("$file.tmp", $file))`

ie. first write both files and then rename them.

src/DI/ContainerLoader.php
```diff
@@ -61,46 +61,57 @@ public function getClassName($key)
...
 		}
 
-		if ($this->autoRebuild) {
-			flock($handle, LOCK_SH);
-			foreach ((array) @unserialize(file_get_contents("$file.meta")) as $f => $time) { // @ - file may not exist
-				if (@filemtime($f) !== $time) { // @ - stat may fail
-					@unlink($file); // @ - file may not exist
-					break;
-				}
+		if (!is_file($file) || $this->isExpired($file)) {
```
dg: Is `isExpired()` here really needed?

JanTvrdik: there is no unlink, so yes

dg: Badly expressed. Is here enough to use previous `isExpered()` result stored in the variable?

JanTvrdik: Depends on what you mean by enough. It will work but it may result in recompiling the container multiple times. The current solution is better.

dg: Yes, I see.

dg:
Merged, little modified 04c7e4d4b016a506ac95955f962cff891101f729

src/DI/ContainerFactory.php
```diff
@@ -110,44 +110,54 @@ private function loadClass()
...
-					throw new Nette\IOException("Unable to create file '$file'.");
-				}
-				$tmp = array();
-				foreach ($this->dependencies as $f) {
-					$tmp[$f] = @filemtime($f); // @ - stat may fail
+		if ((@include $file) === FALSE) { // @ - error escalated to exception
+			throw new Nette\IOException("Unable to include '$file'.");
+		}
+		flock($handle, LOCK_UN);
+	}
```
dg: IMHO lock file can be deleted.

JanTvrdik: IMHO they can't be delete, I think I even tried it and it caused problems. I don't have to retest it or think about it more thoroughly now.

JanTvrdik:
This should have been closed long time ago,
implemented in https://github.com/nette/di/commit/b0fd657c91f02f5acf41ea1ac8feda9057472b1f

PR #39, opened 2015-01-22
Compiler: definition of 'type' will not overide 'factory' of parent service 

dg:
```
services:
     router: 
          type: class
```

will not override parent `factory`


TomasVotruba:
Just came across this on my lecture.

PR #40, opened 2015-01-29
ContainerBuilder: implement escaping of '@' at the beginning of string

xificurk:
Attempt to solve nette/nette#1279

`ContainerBuilder:.formatPhp()` recognizes '\@' at the beginning of string value as escaped '@'.

What do you think about this, is this a reasonable approach? Shall I go ahead and write tests for it?

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -718,7 +718,13 @@ class ContainerBuilder extends Nette\Object
 				$val = '@' . current(array_keys($that->getDefinitions(), $val, TRUE));
 			}
 
-			if (is_string($val) && substr($val, 0, 1) === '@') {
+			if (!is_string($val)) {
+				return;
+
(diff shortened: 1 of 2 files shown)


xificurk:
This for a change prevents passing arguments that should really start with '\@', but I'm not really sure how big of a problem this is in real life.

dg:
I'd like to create solution without escaping, but it will take some time. So it will be solved in 2.4.

dg:
Ok, go ahead and write test ;)

xificurk:
One more idea - using '@@' instead of '\@' would not be a BC break, because strings starting with '@@' never worked in Nette DI.

dg:
`@@` is consistent with `%%` (escaped `%`).

xificurk:
ok, I changed the escape sequence to '@@' and added test.

dg:
Thank you!

fprochazka:
Nice! :)

castamir:
Well done!

PR #42, opened 2015-02-02
Container: optimized getByType() performance

JanTvrdik:


src/DI/Container.php
```diff
@@ -98,8 +98,10 @@ public function removeService($name)
 	 */
 	public function getService($name)
 	{
-		$name = isset($this->meta[self::ALIASES][$name]) ? $this->meta[self::ALIASES][$name] : $name;
 		if (!isset($this->registry[$name])) {
+			if (isset($this->meta[self::ALIASES][$name])) {
+				return $this->getService($this->meta[self::ALIASES][$name]);
+			}
```
dg: Do you think that calling method is faster than accessing array?

JanTvrdik: No

JanTvrdik:
What about this?

src/DI/Container.php
JanTvrdik: The method call could be inlined, do you like it better?

~~~php
public function getService($name)
{
	if (!isset($this->registry[$name])) {
		if (isset($this->meta[self::ALIASES][$name])) {
			$name = $this->meta[self::ALIASES][$name];
			if (!isset($this->registry[$name])) {
				$this->registry[$name] = $this->createService($name);
			}
		} else {
			$this->registry[$name] = $this->createService($name);
		}
	}
	return $this->registry[$name];
}
~~~

dg: It is much more complicated. Did you measure benefit in realworld usage?

JanTvrdik: That I prefer and propose the version if extra `getService` call. It is a simple change which improve `getService` performance in most-common use-case (no aliases).

dg: One lookup in aliases is IMHO immeasurable.

JanTvrdik: But why do it, when you can choose not to do it and still keep the code simple?

dg: You are right that aliases are used very rare.

PR #46, opened 2015-02-03
[WIP] Added Configurator

enumag:
When using DI on its own without some (or all) of the other nette components, one still needs something like `Nette\Configurator`. However `Nette\Configurator` is only available with nette/bootstra which requires lots of other nette/* components and that's not always desired. Therefore some basic Configurator should be added to nette/di. Then nette/bootstrap could just extend it and add what is nette-specific (default extensions, compatibility code, robot loader, tracy).

The only thing I'm not quite sure about is how to call `$this->fixCompatibility($conig)` when needed (https://github.com/nette/bootstrap/blob/master/src/Bootstrap/Configurator.php#L267). I was thinking about using the onCompile event for it at first but the event handlers don't have access to the `$config` array. Any suggestions?

### DIFF
--- a/src/DI/Configurator.php
+++ b/src/DI/Configurator.php
@@ -1,0 +1,279 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2004 David Grudl (http://davidgrudl.com)
+ */
+
+namespace Nette\DI;
+
+use Nette,
+	Tracy;
+
+
+/**
+ * Initial system DI container generator.
+ *
+ * @author     David Grudl
+ *
+ * @property   bool $debugMode
+ * @property-write $tempDirectory
+ */
+class Configurator extends Object
+{
+	const AUTO = TRUE,
+		NONE = FALSE;
+
+	const COOKIE_SECRET = 'nette-debug';
+
+	/** @var callable[]  function(Configurator $sender, Compiler $compiler); Occurs after the compiler is created */
+	public $onCompile;
+
+	/** @var array */
+	public $defaultExtensions = array(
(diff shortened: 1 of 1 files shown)


dg:
You must remove all parameters, detectDebugMode, and defaultExtensions, otherwise it is not a generic configurator.

enumag:
@dg parameters and detectDebugMode are useful pretty much always so I'd like to have such configurator in Nette/DI. Maybe there should be two of them - one completly generic and the other with these basic features.

hrach:
:+1:

dg:
detectDebugMode works with global variables like `$_SERVER['REMOTE_ADDR']`, it has nothing to do with Dependency Injection.

Default parameters exist only for Nette applications, nette/di doesn't require them.

enumag:
@dg Very well. Do you have a suggestion for the issue with fixCompatibility method?

dg:
No

JanTvrdik:
Why don't you just move the `$config` initialization to another method?

enumag:
By the way debugMode is used [here](https://github.com/nette/bootstrap/blob/master/src/Bootstrap/Configurator.php#L217) as well. That can be easily fixed with `public function createContainer($autorebuild = FALSE)` though.

@JanTvrdik That might work, thx. :-)

dg:
I don't understand, your link points to nette/bootstrap.

enumag:
@dg The createContainer method need to be in the generic configurator as well. But it uses debugMode parameter and that needs to be removed.

dg:
Of course.

f3l1x:
I like an idea.

dg:
IMHO this is expired.

enumag:
@dg Yeah. I totally forgot about it. Sorry.

PR #47, opened 2015-02-03
ContainerBuilder: added operator 'raw' as workaround @strings 

dg:
[Closes #40][Closes nette/nette#1279]

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -649,6 +649,9 @@ class ContainerBuilder extends Nette\Object
 		} elseif ($entity === 'not') { // operator
 			return $this->formatPhp('!?', array($arguments[0]));
 
+		} elseif ($entity === 'raw') { // workaround for string that begins with @
+			return PhpHelpers::dump($arguments[0]);
+
 		} elseif (is_string($entity)) { // class name
 			if ($constructor = Reflection\ClassType::from($entity)->getConstructor()) {
 				$this->addDependency($constructor->getFileName());
--- a/tests/DI/Compiler.at-sign.phpt
+++ b/tests/DI/Compiler.at-sign.phpt
@@ -1,0 +1,31 @@
+<?php
+
+/**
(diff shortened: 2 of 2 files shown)


src/DI/ContainerBuilder.php
```diff
@@ -649,6 +649,10 @@ public function formatStatement(Statement $statement)
 		} elseif ($entity === 'not') { // operator
 			return $this->formatPhp('!?', array($arguments[0]));
 
+		} elseif ($entity === 'raw') { // workaround for string that begins with @
+			return PhpHelpers::formatArgs('?', array($arguments[0]));
+			return $this->formatPhp('!?', array($arguments[0]));
```
JanTvrdik: !

src/DI/ContainerBuilder.php
```diff
@@ -649,6 +649,10 @@ public function formatStatement(Statement $statement)
...
+
```
xificurk: eh? return; return; ?

dg: fixed

xificurk:
Am I overlooking something, or does this mean that I would have to use it in `CompilerExtension` as follows?
```php
$arg = '@string.starting.with.at'
$builder->getDefinition('foo')->setArguments(array('name' => new Statement('raw', array($fooArg))));
```

fprochazka:
This is not very nice.

JanTvrdik:
This is not escaping, this is just slightly better workaround than `trim`.

dg:
Hmm, it will not help when you developing CompilerExtension… I'll use you PR.

That strings like `@service` in arguments are parsed as service references was design flaw. It would be better to use any object, for example  `ServiceReference`:

```php
$arg = '@string.starting.with.at'
$builder->getDefinition('foo')->setArguments(array($fooArg, new ServiceReference('bar')));
```

But even we introduce ServiceReference, we still need to parse `@service` for maintain BC…

PR #48, opened 2015-02-05
class names are case sensitive (BC break)

dg:


src/DI/ContainerBuilder.php
```diff
@@ -445,6 +445,18 @@ private function resolveEntityClass($entity, $recursive = array())
 	}
 
 
+	private function checkCase($class)
```
fprochazka: so, this throws an error, when I supply wrongly cased classname in config or into service definition?

dg: or into ContainerBuilder::getByType()

fprochazka: Great :+1:

dg: Uff :-)

JanTvrdik:
:+1: For doing the check in compile-time only.

dg:
Check is done in compile-time only.

JanTvrdik:
I know, that's why I gave you :+1: I didn't realize that the sentence has other meaning as well :smile:

fprochazka:
I just hope I don't have any typos in my extensions :D better check them :)

PR #49, opened 2015-02-05
Remove nette/reflection

dg:


JanTvrdik:
Too aggressive change in RC phase.

dg:
DI is not in RC phase.

JanTvrdik:
I don't see the advantage of rushing partial solution just before release instead of solving the problem properly in next release.

dg:
Nobody told that this must be in 2.3. This is just WIP, I will maybe merge part of it, maybe not. The only last thing that must be solved before 2.3 is https://github.com/nette/di/pull/22

(and maybe API of ContainerLoader, because it is new class)

src/DI/Helpers.php
```diff
@@ -167,4 +168,49 @@ public static function getInjectProperties(Nette\Reflection\ClassType $class, $c
...
 	}
 
+
+	/**
+	 * Returns an annotation value.
+	 * @return string|NULL
+	 */
+	public static function getAnnotation(\Reflector $ref, $name)
+	{
+		if (preg_match("#\\s@$name(?:\\s++([^@]\\S*))?#", $ref->getDocComment(), $m)) {
```
fprochazka: I've never seen  `++` in any regexp, what is it good for?

fprochazka:
@dg why haven't the travis tested this pullrq? 

@JanTvrdik I don't consider this a BC Break or rushed. If enough time is given for RC to test this properly, I have no problem with this beeing in 2.3.

fprochazka:
@dg you might also wanna throw away dependency on `nette/reflection` from `composer.json`

dg:
It is not ready, it still depends on nette/reflection. 

Why it is not tested? I have no idea. Travis works, but Github ignores it.

src/DI/Helpers.php
JanTvrdik: prevents backtracking

enumag: @fprochazka See http://www.regular-expressions.info/possessive.html. I use these quite often.

JanTvrdik:
@fprochazka To do it properly you need to consider all packages depending on nette/reflection and probably rewrite annotation support from scratch and put it to a branch new package. Or maybe not. I don't. Noone knows and there is a little time to find the best possible solution at this point.

dg:
Now it is complete and ready. 

There is one difference to nette/reflection: it doesn't cache. Intentionally. 

Cache was originally used for compatiblity with opcode caches that discard annotations (phpDoc comments). But yesterday I realized that Nette since 2.1 is not using this fallback for ObjectMixin, so it is not possible to run Nette since 2.1 when annotations are not available. So I merged nette/application#55, it is not BC break.

Cache was used for parsing of PHP files to analyse use statement (btw: sad story https://wiki.php.net/rfc/aliases_by_reflection). It was effective only when you used `@inject` and type was not fully qualified and presenter was not in DIC. 

Nette 2.3 is trying to add all presenters to DIC and `@inject` is optional feature, so I am not sure whether it is needed to implement cache here too or if it should trigger warning…

fprochazka:
As it's ran only on compile-time, I think the cache is unnecessary.

I extremely dislike the code duplication. Is it worth it, in the age of composer, to duplicate so much code?

fprochazka:
I'd be willing to overlook the yet another awful copy&paste, if the code is needed only inside `nette/di` and nowhere else. But I'm not sure that this is the case.

enumag:
In my opinion the reflection helpers can be useful for other people as well. It would be better to not kill nette/reflection completely, just reduce it to such static helpers and deprecate (later remove) everything else from that package.

JanTvrdik:
I still think that the problem is too complex to be solved in such short time.

dg:
I'd like to abandod nette/reflection. The only duplication here is https://github.com/nette/application/blob/master/src/Application/UI/PresenterComponentReflection.php#L170 and https://github.com/dg/nette-di/blob/5935c2460afa1b7523c17035d0e04ef98efaf29a/src/DI/PhpReflection.php#L29, nothing else.

The benefit of small repositories is that it is not needed to deprecate it or change it with BC breaks. It will be simply not used by other parts of Nette anymore.

enumag:
@dg Put the helpers to nette/utils then.

enumag:
@dg Also about the code duplication, even if there isn't much of code duplication in Nette itself, by marking it as internal you're forcing other people to duplicate it.

dg:
Maybe in future. But `getAnnotation` is not good solution, it is quick'n'dirty. And `getDeclaringClass()`, `getPropertyType()` and `expandClassName()` are very rarely used.

JanTvrdik:
> I'd like to abandod nette/reflection. 
[So do I](http://forum.nette.org/en/20766-nette-hhvm-how-is-the-hhvm-support-status#p142736), I just think that it should not be rushed.

dg:
What is rushed? nette/reflection 2.3 is ready. Code in nette/application and nette/di is simple and will be tested in next few weeks. I plan to abandon nette/reflection halt a year. What is rushed?

enumag:
@dg Well I actually do use `expandClassName()` in one of my private package and after looking at `getPropertyType()` I want to use it as well so thank you very much. ;-)

JanTvrdik:
> will be tested in next few weeks

The testing period should be primary used for fixing bugs in implementation not for verifying design decisions. Design decisions usually require much longer verification period (e.g. couple of months). But do as you please I don't have time to discuss this with you any further. Hopefully it will have happy handing.

dg:
For me this is not design decision, only some functions was moved from one repository to new internal (!) class. It affects nearly nothing, API is the same. In n/application, an internal class PresenterComponentReflection is still descendant of ReflectionClass, but not Reflection\ClassType. Who cares. Code is cleaner, one dependency shorter. This change perfectly fits for minor version. 

Of couse, some changes require long time of testing. I have long list of features that I forbid to myself to merge them to 2.3, because they are the "design decisions". Replacing nette/reflection with native one classes was easy pick, routine, internal cleanup.

dg:
But maybe I don't understand you, why do think it affects design?

fprochazka:
I don't think that it is a design decision.

Btw, what about [afterCompile](https://github.com/nette/di/blob/a9964ebf4420d9598fc35b4066a1d54cda929f71/src/DI/CompilerExtension.php#L135)? Unless you remove the dependency, `nette/di` is still dependent on `nette/reflection`. But removing it means **one hell of a fucking nasty bc break**. Changing typehints are the works kind of bc breaks. I would be so pissed if I had to maintain another release branch on all my repos just because of this one line.

PR #50, opened 2015-02-06
CompilerExtension - getConfig: fixed BC break from @181e141b

matej21:
see comment: https://github.com/nette/di/commit/181e141b

### DIFF
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -46,6 +46,11 @@ class Compiler extends Nette\Object
 		if (isset(self::$reserved[$name])) {
 			throw new Nette\InvalidArgumentException("Name '$name' is reserved.");
 		}
+		if (isset($this->config[$name])) {
+			$this->config[$name] = $tmp = Helpers::expand($this->config[$name], $this->builder->parameters);
+			unset($tmp['services']);
+			$extension->setConfig($tmp);
+		}
 		$this->extensions[$name] = $extension->setCompiler($this, $name);
 		return $this;
 	}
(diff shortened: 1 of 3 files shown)


dg:
What about change the Compiler that can in first round call setConfig on all extensions and then in second round loadConfiguration?

It would be necessary still call setConfig for each newly added extension.

matej21:
Ok, I added alternative solution (and updated test case with edge case). Choose one and I squash it :)

src/DI/Compiler.php
```diff
@@ -106,13 +111,15 @@ public function processParameters()
...
-			$name = key($slice);
+		foreach ($this->extensions as $name => $extension) {
 			if (isset($this->config[$name])) {
 				$this->config[$name] = $tmp = Helpers::expand($this->config[$name], $this->builder->parameters);
 				unset($tmp['services']);
 				$this->extensions[$name]->setConfig($tmp);
 			}
+		}
+		for ($i = 0; $slice = array_slice($this->extensions, $i, 1, TRUE); $i++) {
+			$name = key($slice);
```
dg: $what about something like 

```php
$last = count($this->extensions);
for ($i = 0; $slice = array_slice($this->extensions, $i, 1, TRUE); $i++) {
      $name = key($slice);
      if ($i >= $last)  /* setConfig */
      $this->extensions[$name]->loadConfiguration();
 }
```

matej21: This is a replacement of the code in addExtension? I'm not sure I understand you, but I think it cannot work :)

dg: Yes, this should call setConfig in loop in processParameters() instead of addExtension.

matej21: In processParameters? Now I am totally lost :) For new extensions (which added by other extension) we need to call setConfig immediately after extension is added (see test - Foo adds Bar extension and calls getConfig...)

dg:
Sry, processExtension

src/DI/Compiler.php
dg: Sry, processExtension

matej21: It will work with most use cases. 
But it won't work with some edge cases - e.g. the test bellow. Config must be set right after [extension is added](https://github.com/nette/di/pull/50/files#diff-eaabdd91a6dedf0fa69456f1132d3483R22). Second edge case I can think of is when extension replaces other extension.

dg:
I feel that all this dynamic adding/changing extensions is design flaw…

Is there any reason to add/change other extensions except what does ExtensionsExtension? Is not, I am for simplification: ExtensionsExtension will be called as first, this patch will not be needed and as loop in processExtensions can be used foreach.

We really need to solve problem with extension order (task for 2.4) and to remove unnecessary features seems like a good start.

matej21:
@dg ok, I agree. 
But I think getConfig in CompilerExtension should at least throw some exception if config has not been set yet.

fprochazka:
@dg I have no problem with that. But there are AFAIK people, that try to hack nette/di for something "more modular" and are adding extensions in their "ModulesCoreExtensions".

enumag:
@dg :+1:

dg:
Any extension can add other extensions like *broker*, the same way how RouteList works, but changing the global extensions namespace is IMHO bad idea. This is space for end user, for his config.neon.

PR #62, opened 2015-02-27
Improved Compiler & Loader extendability

lookyman:


JanTvrdik:
Can you describe the use case?

lookyman:
```php
<?php

use Tester\Assert;

require __DIR__ . '/../bootstrap.php';

class TestCompiler extends Nette\DI\Compiler
{
	public function createLoader()
	{
		return (new Nette\DI\Config\Loader)->addAdapter('yml', new YmlAdapter);
	}
}

$compiler = new TestCompiler;
$compiler->loadConfig(__DIR__ . '/config.yml');

// Tester::assert(...)
```

And yes, I do realize that currently I have to override a method that is only three lines longer..

dg:
createLoader & createCompiler should be protected.

lookyman:
@dg right..

TomasVotruba:
This looks ready @dg 

Anything else that needs to be done here?

Majkl578:
It needs rebase at least...

TomasVotruba:
Thank you @lookyman

lookyman:
Hmm, coveralls is failing. I can look into writing more tests over the weekend..

TomasVotruba:
I think coveralls is passing, just unhappy about decreasing the coverage.

lookyman:
Ok, added a quick test to make coveralls happy.

TomasVotruba:
Wow, you are awesome. Thanks!

f3l1x:
:+1:

TomasVotruba:
I wonder, correct place to allow extension using constructor injection would be:

```php
public function __construct(ILoader $loader = NULL)
{
	$this->loader = $loader ?: new Loader;
}
```

dg:
To inject different Loader to compiler extension has no sense. Or did you mean a different class?

TomasVotruba:
@dg Do you react to me?
If so,  I mean the same change just via dependency not via child override.

dg:
ad custom loader in Compiler::loadConfig()
- it should be set via code that uses class, no via class itself. Because class can't know what you will pass to loadConfig()
- so IMHO best way is Compiler::loadConfig($file, Loader $loader = NULL) 

ad custom loader in CompilerExtension::loadFromFile()
- CompilerExtension descendant (in contrast with Compiler) knows what is passed to loadFromFile(), so createLoader() is fine

ad custom compiler in ContainerLoader::createCompiler()
- $generator passed to load() allows you to override $compiler, i.e current $compiler is only default value. And it is preferred way, so createCompiler() is not needed.

PR #65, opened 2015-03-11
Fine grained dependencies 

dg:


dg:
Can you test it before merge?

PR #71, opened 2015-06-02
ContainerBuilder: removeDefinition removes from classes too

TomasVotruba:
Use case:

```php
$containerBuilder->addDefintion('one')->setClass('stdClass');
$containerBuilder->addDefintion('two')->setClass('stdClass');
$containerBuilder->removeDefinition('two');
$containerBuilder->getByType('stdClass'); // breaks, because reference to 'two' is still saved under 'stdClass'
```

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -81,12 +81,12 @@ class ContainerBuilder extends Nette\Object
 
 		if (isset($this->definitions[$name])) {
 			$class = $this->definitions[$name]->getClass();
-			if (isset($this->classes[$class][TRUE])) {
-				foreach ($this->classes[$class][TRUE] as $key => $definitionName) {
-					if ($name === $definitionName) {
-						unset($this->classes[$class][TRUE][$key]);
-					}
-				}
+			$this->removeClassFromReference($name, $class);
+
+			$reflectionClass = new ReflectionClass($class);
+			while ($parentClass = $reflectionClass->getParentClass()) {
+				$this->removeClassFromReference($name, $parentClass->getName());
+				$reflectionClass = new ReflectionClass($parentClass->getName());
 			}
 
 			unset($this->definitions[$name]);
@@ -870,6 +870,24 @@ class ContainerBuilder extends Nette\Object
 			throw new ServiceCreationException("Reference to missing service '$service'.");
 		}
 		return $service;
+	}
+
+
+	/**
+	 * @param string $name
+	 * @param string $class
+	 */
+	private function removeClassFromReference($name, $class)
+	{
+		foreach ([TRUE, FALSE] as $mode) {
+			if (isset($this->classes[$class][$mode])) {
+				foreach ($this->classes[$class][$mode] as $key => $definitionName) {
+					if ($name === $definitionName) {
+						unset($this->classes[$class][$mode][$key]);
+					}
+				}
+			}
+		}
 	}
 
 }
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -78,7 +78,19 @@ class ContainerBuilder extends Nette\Object
 	public function removeDefinition($name)
 	{
 		$name = isset($this->aliases[$name]) ? $this->aliases[$name] : $name;
-		unset($this->definitions[$name]);
+
+		if (isset($this->definitions[$name])) {
+			$class = $this->definitions[$name]->getClass();
+			if (isset($this->classes[$class][TRUE])) {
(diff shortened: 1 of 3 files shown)


fprochazka:
Nice addition. If you could also add tests, that would be awesome.

TomasVotruba:
I'm on it.

TomasVotruba:
It's ready now.

fprochazka:
nice :+1:

src/DI/ContainerBuilder.php
```diff
@@ -78,7 +78,19 @@ public function addDefinition($name, ServiceDefinition $definition = NULL)
 	public function removeDefinition($name)
 	{
 		$name = isset($this->aliases[$name]) ? $this->aliases[$name] : $name;
-		unset($this->definitions[$name]);
+
+		if (isset($this->definitions[$name])) {
+			$class = $this->definitions[$name]->getClass();
+			if (isset($this->classes[$class][TRUE])) {
+				foreach ($this->classes[$class][TRUE] as $key => $definitionName) {
```
dg: What about `$this->classes[$class][FALSE]`?

TomasVotruba:
@dg Added!

src/DI/ContainerBuilder.php
TomasVotruba: @dg Added!

dg:
All occurrences should be removed, not only in `$this->classes[$class]`.

TomasVotruba:
@dg What do you mean?

dg:
for example `$this->classes[get_parent_class($class)]` etc.

TomasVotruba:
@dg I still don't undestand the point. I'm not familiar with $classes content in such depth. Could you provide failing usecase?

dg:
class A extends stdClass {}

$containerBuilder->addDefintion('two')->setClass('A');

TomasVotruba:
This might do it. Please check

dg:
```
foreach ($this->classes as $class => $tmp) 
   foreach ($tmp as $mode => $nm) {
     if ($name = $nm) unset($this->classes[$class][$tmp])
}

File: src/DI/ContainerBuilder.php [hunk 2]
```diff
@@ -78,7 +78,22 @@ public function addDefinition($name, ServiceDefinition $definition = NULL)
 	public function removeDefinition($name)
 	{
 		$name = isset($this->aliases[$name]) ? $this->aliases[$name] : $name;
-		unset($this->definitions[$name]);
+
+		if (isset($this->definitions[$name])) {
+			if ($this->classes) {
+				foreach ($this->classes as $class => $tmp) {
+					foreach ($tmp as $mode => $serviceNames) {
+						foreach ($serviceNames as $key => $serviceName) {
```
dg: Hmmm, and what about this? `$this->classes[$class][$mode] = array_diff($this->classes[$class][$mode], $name)`.

src/DI/ContainerBuilder.php
```diff
@@ -78,7 +78,22 @@ public function addDefinition($name, ServiceDefinition $definition = NULL)
 	public function removeDefinition($name)
 	{
 		$name = isset($this->aliases[$name]) ? $this->aliases[$name] : $name;
-		unset($this->definitions[$name]);
+
+		if (isset($this->definitions[$name])) {
```
dg: if (isset($this->definitions[$name]) && $this->classes) {

src/DI/ContainerBuilder.php
TomasVotruba: I thinks it less readable. It might be faster, container compilation is cached though.

src/DI/ContainerBuilder.php
```diff
@@ -860,4 +875,22 @@ public function getServiceName($arg)
 		return $service;
 	}
 
+
+	/**
+	 * @param string $name
+	 * @param string $class
+	 */
+	private function removeClassFromReference($name, $class)
+	{
```
dg: ?

TomasVotruba: Dev code, I shall removed it.

TomasVotruba:
Parfe!

dg:
Thanx

TomasVotruba:
One more bug:

```php
$someInterface => [
    0 => FirstService
    1 => SecondService
];
```

Remove 'FirstService':


```php
$someInterface => [
    1 => SecondService
];
```

[Then, `getByType` will try to get [0]](https://github.com/nette/di/blob/72bc3f0ade6d9c8f0e302e87466cd5c291ffac24/src/DI/ContainerBuilder.php#L207) => FAIL

Solution? Something like:

```php
return reset($this->classes[$class][TRUE][0]);
```

What do you thing?

PR #72, opened 2015-06-09
removeDefinition fix for removing first added one

TomasVotruba:
One more related bug to #71:

```php
$someInterface => [
    0 => FirstService
    1 => SecondService
];
```

Remove 'FirstService':

```php
$someInterface => [
    1 => SecondService
];
```

[Then, `getByType` will try to get [0]](https://github.com/nette/di/blob/72bc3f0ade6d9c8f0e302e87466cd5c291ffac24/src/DI/ContainerBuilder.php#L207) => FAIL

Solution? Something like:

```php
return reset($this->classes[$class][TRUE]);
```

What do you think?


TomasVotruba:
:+1:

PR #76, opened 2015-07-12
Xml adapter

sallyx:
# Xml Adapter
XML adapter could be used to write configuration in XML. 
Thanks to namespace usage (http://www.nette.org/xmlns/nette/config/1.0) you can have  configuration common for nette and other application, configuration for nette  only and  for other application only in the same file.

For basic usage example: see [tests/DI/files/xmlAdapter.xml](tests/DI/files/xmlAdapter.xml)

### Associative array 

```xml
<myarray><key1>value1</key1><key2>value2</key2></myarray>
```
Becomes:
```php
[key1 => 'value1', key2=>'value2']
```
### Numeric array

```xml
<myarray array="numeric"><key>value1</key><key>value2</key></myarray>
```
Becomes (*element names are ignored here*):
```php
[0=>'value1',1=>'value2]
```
### String, Null, Number, Boolean

```xml
<myarray array="numeric">
<x> trimmed string </x>
<x space="preserve"> string with spaces  </x>
<x>1</x>
<x number="2"/>
<x bool="yes"/>
<x null="null" />
</myarray>
```
Becomes:
```php
["trimmed string", " string with spaces  ","1", 2, TRUE, NULL]
```
*Bool support "yes", "true", "on" and "1"*
### Statement
```xml
<factory statement="statement">
    <s><ent>DateTime</ent><args><a numeric="0" /></args></s>
   <s><ent>format</ent><args><a>%B</a></args></s>
</factory>
```
Is equivalent to neon: ```DateTime(0)::format("%B")```

### Array from string
```xml
<myarray array="string">1,2,3</myarray>
```
Becomes:
```php
["1", "2", "3"]
```
#### The same with custom delimiter:
```xml
<myarray array="string" delimiter=";">1;2;3</myarray>
```

### Statement
Of course you can use array from string in statement arguments.
```xml
<xxx statement="statement">
<s><ent>fooo</ent><args array="string" delimiter=";">1;2;3</args></s>
</xxx>
```
If there is only one argument, you dont need to use element inside args element.
These statements are equivalent:
```xml
<xxx statement="statement"><s><ent>fooo</ent><args><a>1</a></args></s></xxx>
<xxx statement="statement"><s><ent>fooo</ent><args>1</args></s></xxx>
```

You can use neon syntax for statement:
```xml
<xxx statement="DateTime(0)::format('%B')" />
```

## Example of converting neon config file to xml config file

```php
use Nette\DI\Config\Adapters\NeonAdapter;
use Nette\DI\Config\Adapters\XmlAdapter;

$na = NeonAdapter;
$xa = XmlAdapter;
$config = $na->load('config.neon');
$xmlConfig = $xmlloader->dump($config);
// pretty output
$domxml = new DOMDocument('1.0');
$domxml->preserveWhiteSpace = false;
$domxml->formatOutput = true;
$domxml->loadXML($xmlConfig);
echo $domxml->saveXML();
```

### DIFF
--- a/src/DI/Config/Adapters/XmlAdapter.php
+++ b/src/DI/Config/Adapters/XmlAdapter.php
@@ -1,0 +1,427 @@
+<?php
+
+/**
+ * This file is part of the Nette Framework (http://nette.org)
+ * Copyright (c) 2015 Petr Bilek (http://ww.sallyx.org)
+ */
+
+namespace Nette\DI\Config\Adapters;
+
+use Nette;
+use Nette\DI\Config\Helpers;
+use Nette\DI\Statement;
+
+/**
+ * Reading and generating XML files.
+ */
+class XmlAdapter extends Nette\Object implements Nette\DI\Config\IAdapter
+{
+
+	const NS = "http://www.nette.org/xmlns/nette/config/1.0";
+
+	/**
+	 * Reads configuration from XML file.
+	 * @param  string  file name
+	 * @return array
+	 */
+	public function load($file)
+	{
+		$opt = LIBXML_NOBLANKS | LIBXML_NOCDATA | LIBXML_NOENT | LIBXML_NSCLEAN; // | LIBXML_PEDANTIC ?
+		$parserClass = "\\Nette\\DI\\Config\\Adapters\\XMLElementParser";
+		$doc = simplexml_load_file($file, $parserClass, $opt, self::NS);
+		return $doc->parse();
+	}
+
+	/**
+	 * Generates configuration in XML format.
+	 * @return string
+	 */
+	public function dump(array $data)
+	{
+		$writter = new XmlElementWritter('<config/>', 0, false, self::NS, "nc");
+		$writter->addAttribute('xmlns:xmlns:nc', self::NS);
+		$writter->addAttribute('xmlns', self::NS);
+		$writter->addConfig($data);
+		return $writter->asXML();
+	}
+
+}
+
+/**
+ * This class helps to parse XML config to PHP array
+ * @internal
+ */
+class XMLElementParser extends \SimpleXMLElement
+{
+
+	/** Names of attributes */
+	const ATTR_ARRAY = 'array',
+		ATTR_BOOL = 'bool',
+		ATTR_NUMBER = 'number',
+		ATTR_DELIMITER = 'delimiter',
+		ATTR_EXTENDS = 'extends',
+		ATTR_MERGING = 'merging',
+		ATTR_NULL = 'null',
+		ATTR_STATEMENT = 'statement',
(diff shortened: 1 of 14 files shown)


src/DI/Config/Adapters/XmlAdapter.php
```diff
@@ -0,0 +1,427 @@
...
+	const NS = "http://www.nette.org/xmlns/nette/config/1.0";
+
+	/**
+	 * Reads configuration from XML file.
+	 * @param  string  file name
+	 * @return array
+	 */
+	public function load($file)
+	{
+		$opt = LIBXML_NOBLANKS | LIBXML_NOCDATA | LIBXML_NOENT | LIBXML_NSCLEAN; // | LIBXML_PEDANTIC ?
```
lookyman: `$options`

src/DI/Config/Adapters/XmlAdapter.php
```diff
@@ -0,0 +1,427 @@
...
+		$parserClass = "\\Nette\\DI\\Config\\Adapters\\XMLElementParser";
+		$doc = simplexml_load_file($file, $parserClass, $opt, self::NS);
```
lookyman: `$document`

src/DI/Config/Adapters/XmlAdapter.php
```diff
@@ -0,0 +1,427 @@
...
+		return $doc->parse();
+	}
+
+	/**
+	 * Generates configuration in XML format.
+	 * @return string
+	 */
+	public function dump(array $data)
+	{
+		$writter = new XmlElementWritter('<config/>', LIBXML_NOEMPTYTAG, false, self::NS, "nc");
```
lookyman: `$writer`

src/DI/Config/Adapters/XmlAdapter.php
```diff
@@ -0,0 +1,427 @@
...
+		return explode($delimiter[0], $value);
+	}
+
+}
+
+/**
+ * This class helps to save array into XML configuration
+ * @internal
+ */
+class XMLElementWritter extends \SimpleXMLElement
```
lookyman: `XmlElementWriter`

dg:
I don't like to add this adapter to Nette\DI, so move it to your own package please.

sallyx:
Why not?

dg:
Because (nearly) nobody needs it and I don't want to maintain the code.

fprochazka:
@sallyx is there anything that prevents you from adding the xml support as an external package? Having it separate is not so bad idea.

sallyx:
@fprochazka Yes. I do not know how to do it :smiley_cat:

dg:
Thanks @sallyx for PR, but this is really rare case, so please publish it as addon.

PR #77, opened 2015-08-19
Support for PHP7 return types and scalar parameters for factories

pavelkouril:
**Depends on https://github.com/nette/php-generator/pull/11**

- feature
- documentation - not needed
- BC break - no

Adds support for PHP 7 syntax and features:
- factories can now use return type declarations instead of the `@return` annotation
- scalar type hints can be used in parameters of the factory interfaces

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -421,7 +421,7 @@ class ContainerBuilder extends Nette\Object
 		if ($class = $def->getClass() ?: $class) {
 			$def->setClass($class);
 			if (!class_exists($class) && !interface_exists($class)) {
-				throw new ServiceCreationException("Class or interface $class used in service '$name' not found.");
+				throw new ServiceCreationException("Type $class used in service '$name' not found.");
 			}
 			self::checkCase($class);
 
@@ -638,7 +638,8 @@ class ContainerBuilder extends Nette\Object
 
 		$factoryClass->addMethod($def->getImplementType())
 			->setParameters($this->convertParameters($def->parameters))
-			->setBody(str_replace('$this', '$this->container', $code));
+			->setBody(str_replace('$this', '$this->container', $code))
+			->setReturnType(PHP_VERSION_ID >= 70000 ? $def->getClass() : NULL);
 
 		return "return new {$factoryClass->getName()}(\$this);";
 	}
--- a/src/DI/PhpReflection.php
+++ b/src/DI/PhpReflection.php
@@ -59,7 +59,9 @@ class PhpReflection
 	 */
 	public static function getParameterType(\ReflectionParameter $param)
 	{
-		if ($param->isArray() || $param->isCallable()) {
+		if (PHP_VERSION_ID >= 70000 && $param->hasType()) {
+			return (string) $param->getType();
+		} elseif ($param->isArray() || $param->isCallable()) {
 			return $param->isArray() ? 'array' : 'callable';
 		} else {
 			try {
@@ -79,6 +81,9 @@ class PhpReflection
 	 */
 	public static function getReturnType(\ReflectionFunctionAbstract $func)
 	{
+		if (PHP_VERSION_ID >= 70000 && $func->hasReturnType()) {
+			return (string) $func->getReturnType();
+		}
 		$type = preg_replace('#[|\s].*#', '', (string) self::parseAnnotation($func, 'return'));
 		if ($type) {
 			return $func instanceof \ReflectionMethod
--- a/tests/DI/Compiler.generatedFactory.returnTypes.phpt
+++ b/tests/DI/Compiler.generatedFactory.returnTypes.phpt
@@ -1,0 +1,42 @@
+<?php
+
+/**
+ * Test: Nette\DI\Compiler: generated services factories from interfaces with return type declarations.
+ * @phpVersion 7.0
+ */
+
+use Nette\DI;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+interface IArticleFactory
+{
+
+	function create($title): Article;
+}
+
+class Article
+{
+	public $title;
+
+	function __construct($title)
+	{
+		$this->title = $title;
+	}
+}
+
+$compiler = new DI\Compiler;
+$container = createContainer($compiler, 'files/compiler.generatedFactory.returnTypes.neon');
+
+Assert::type('IArticleFactory', $container->getService('article'));
+$article = $container->getService('article')->create('lorem-ipsum');
+Assert::type('Article', $article);
+Assert::same('lorem-ipsum', $article->title);
+
+Assert::type('IArticleFactory', $container->getService('article2'));
+$article = $container->getService('article2')->create('lorem-ipsum');
+Assert::type('Article', $article);
+Assert::same('lorem-ipsum', $article->title);
--- a/tests/DI/Compiler.generatedFactory.scalarParameters.phpt
+++ b/tests/DI/Compiler.generatedFactory.scalarParameters.phpt
@@ -1,0 +1,48 @@
+<?php
+
+/**
+ * Test: Nette\DI\Compiler: generated services factories from interfaces with scalar type hints in parameters.
+ * @phpVersion 7.0
+ */
+
+use Nette\DI;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+interface IArticleFactory
+{
+
+	/** @return Article */
+	function create(string $title);
+}
+
+class Article
+{
+	public $title;
+
+	function __construct(string $title)
+	{
+		$this->title = $title;
+	}
+}
+
+$compiler = new DI\Compiler;
+$container = createContainer($compiler, 'files/compiler.generatedFactory.scalarParameters.neon');
+
+Assert::type('IArticleFactory', $container->getService('article'));
+$article = $container->getService('article')->create('lorem-ipsum');
+Assert::type('Article', $article);
+Assert::same('lorem-ipsum', $article->title);
+
+Assert::type('IArticleFactory', $container->getService('article2'));
+$article = $container->getService('article2')->create('lorem-ipsum');
+Assert::type('Article', $article);
+Assert::same('lorem-ipsum', $article->title);
+
+Assert::type('IArticleFactory', $container->getService('article3'));
+$article = $container->getService('article3')->create('lorem-ipsum');
+Assert::type('Article', $article);
+Assert::same('lorem-ipsum', $article->title);
--- a/tests/DI/ContainerBuilder.factory.resolveBuiltinTypes.php5.phpt
+++ b/tests/DI/ContainerBuilder.factory.resolveBuiltinTypes.php5.phpt
@@ -1,0 +1,53 @@
+<?php
+
+/**
+ * Test: Nette\DI\ContainerBuilder and resolving builtin types for generated factories.
+ */
+
+namespace A
+{
+
+	class Factory
+	{
+		/** @return array */
+		function createArray()
+		{
+			return [];
+		}
+
+		/** @return callable */
+		function createCallable()
+		{
+			return function () {};
+		}
+	}
+
+}
+
+namespace
+{
+	use Nette\DI;
+	use Tester\Assert;
+
+
+	require __DIR__ . '/../bootstrap.php';
+
+	Assert::exception(function () {
+		$builder = new DI\ContainerBuilder;
+		$builder->addDefinition('factory')
+			->setClass('A\Factory');
+		$builder->addDefinition('a')
+			->setFactory('@factory::createArray');
+		$container = createContainer($builder);
+	}, Nette\DI\ServiceCreationException::class, "Type array used in service 'a' not found.");
+
+	Assert::exception(function () {
+		$builder = new DI\ContainerBuilder;
+		$builder->addDefinition('factory')
+			->setClass('A\Factory');
+		$builder->addDefinition('c')
+			->setFactory('@factory::createCallable');
+		$container = createContainer($builder);
+	}, Nette\DI\ServiceCreationException::class, "Type callable used in service 'c' not found.");
+
+}
--- a/tests/DI/ContainerBuilder.factory.resolveBuiltinTypes.php7.phpt
+++ b/tests/DI/ContainerBuilder.factory.resolveBuiltinTypes.php7.phpt
@@ -1,0 +1,83 @@
+<?php
+
+/**
+ * Test: Nette\DI\ContainerBuilder and resolving builtin types for generated factories. Added checks for types added in PHP 7.0.
+ * @phpVersion 7.0
+ */
+
+namespace A
+{
+
+	class Factory
+	{
+		/** @return string */
+		function createString()
+		{
+			return "";
+		}
+
+		/** @return int */
+		function createInt()
+		{
+			return 0;
+		}
+
+		function createBool(): bool
+		{
+			return FALSE;
+		}
+
+		function createFloat(): float
+		{
+			return 0.0;
+		}
+	}
+
+}
+
+namespace
+{
+	use Nette\DI;
+	use Tester\Assert;
+
+
+	require __DIR__ . '/../bootstrap.php';
+
+
+	Assert::exception(function () {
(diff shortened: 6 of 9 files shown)


src/DI/ContainerBuilder.php
```diff
@@ -411,6 +417,17 @@ private function resolveImplement(ServiceDefinition $def, $name)
 	}
 
 
+	private function getTypeHintForParameter(\ReflectionParameter $param)
+	{
+		if (PHP_VERSION_ID >= 70000 && $param->hasType()) {
+			$type = $param->getType();
+			return $type->isBuiltin() ? (string) $type : '\\' . $type;
+		} else {
+			return $param->isArray() ? 'array' : PhpReflection::getPropertyType($param);
```
enumag: What about callable?

src/DI/ContainerBuilder.php
```diff
@@ -411,6 +417,17 @@ private function resolveImplement(ServiceDefinition $def, $name)
 	}
 
 
+	private function getTypeHintForParameter(\ReflectionParameter $param)
+	{
+		if (PHP_VERSION_ID >= 70000 && $param->hasType()) {
+			$type = $param->getType();
+			return $type->isBuiltin() ? (string) $type : '\\' . $type;
+		} else {
```
Majkl578: No need for else.

src/DI/ContainerBuilder.php
pavelkouril: What about it? Callable is considered a built-in type by `ReflectionType::isBuiltin`.

enumag: That is in the condition for PHP 7. Will it work for PHP 5.4 - 5.6?

pavelkouril: Functionality for 5.x stays the same as it was before; the else branch is [exactly the same as it used to be](https://github.com/nette/di/blob/master/src/DI/ContainerBuilder.php#L395). I'm honestly not sure if PR for PHP7 features should change the functionality on PHP 5.x.

What do you think?

enumag: Yeah probably not... although it probably doesn't work now...

dg: Probably support for callable should be merged before this PR. This also applies to nette/php-generator.

src/DI/ContainerBuilder.php
```diff
@@ -355,12 +355,18 @@ private function resolveImplement(ServiceDefinition $def, $name)
 		$def->setImplementType($methodName = $rc->hasMethod('create') ? 'create' : 'get');
 
 		if (!$def->getClass() && !$def->getEntity()) {
-			$returnType = PhpReflection::parseAnnotation($method, 'return');
-			if (!$returnType) {
-				throw new ServiceCreationException("Method $interface::$methodName() used in service '$name' has no @return annotation.");
+			if (PHP_VERSION_ID >= 70000 && $method->hasReturnType()) {
+				$type = $method->getReturnType();
+				$returnType = $type->isBuiltin() ? (string) $type : '\\' . $type;
```
dg: `$returnType = $method->getReturnType();` is IMHO enough

src/DI/ContainerBuilder.php
```diff
@@ -411,6 +417,16 @@ private function resolveImplement(ServiceDefinition $def, $name)
 	}
 
 
+	private function getTypeHintForParameter(\ReflectionParameter $param)
+	{
+		if (PHP_VERSION_ID >= 70000 && $param->hasType()) {
+			$type = $param->getType();
+			return $type->isBuiltin() ? (string)$type : '\\' . $type;
```
dg: `return $param->getType();` is IMHO enough. And probably functionality of getTypeHintForParameter() can be moved to `PhpReflection::getPropertyType()`.

src/DI/ContainerBuilder.php
```diff
@@ -643,9 +664,12 @@ private function generateService($name)
 			->addParameter('container')
 				->setTypeHint($this->className);
 
-		$factoryClass->addMethod($def->getImplementType())
+		$method = $factoryClass->addMethod($def->getImplementType())
 			->setParameters($this->convertParameters($def->parameters))
 			->setBody(str_replace('$this', '$this->container', $code));
+		if (PHP_VERSION_ID >= 70000) {
```
dg: What about `->setReturnType(PHP_VERSION_ID >= 70000 ? $def->getClass() : NULL);` Just idea…

pavelkouril: Yeah, this looks way better. :) There is one small problem with the return type adding which was pointed out to me on chat yesterday - it currently adds the return type declaration even if it's not specified in interface. 

It doesn't break applications (and tests) because PHP allows adding the return type in any point of hierarchy, but still changes how the FW behaved up until now - is that considered a BC break (and a issue in general), since it affects only automatically generated classes and doesn't break anything?

src/DI/ContainerBuilder.php
pavelkouril: Would it be a problem if the callable support would get merged after/in this commit after all? I can do it (actually have it already, including test), but I built it on this PR - so it would be easier to get it accepted after (so I don't have to rebase this PR).

Or we can add it in this PR and rename it to "add support for callable types and PHP 7 type hints" instead of just "add support for PHP7"?

But if you insist on doing first the callable and this, I can create another PR and then rebase this one - it's up to you, I'm not familiar with netter PR process that much. :)

src/DI/PhpReflection.php
```diff
@@ -59,13 +59,20 @@ public static function getDeclaringClass(\ReflectionProperty $prop)
...
 	public static function getPropertyType(\ReflectionParameter $prop)
 	{
-		try {
-			return ($ref = $prop->getClass()) ? $ref->getName() : NULL;
-		} catch (\ReflectionException $e) {
-			if (preg_match('#Class (.+) does not exist#', $e->getMessage(), $m)) {
-				return $m[1];
+		if (PHP_VERSION_ID >= 70000) {
+			$type = $prop->getType();
+			return !$type || $type->isBuiltin() ? (string) $type : '\\' . $type;
```
dg: Why is here `'\\' . $type`?

src/DI/ContainerBuilder.php
```diff
@@ -463,9 +468,14 @@ private function resolveEntityClass($entity, $recursive = [])
 				throw new ServiceCreationException(sprintf("Factory '%s' used in service '%s' is not callable.", Nette\Utils\Callback::toString($entity), $name[0]));
 			}
 
-			$class = preg_replace('#[|\s].*#', '', PhpReflection::parseAnnotation($reflection, 'return'));
-			if ($class) {
-				$class = $refClass ? PhpReflection::expandClassName($class, $refClass) : ltrim($class, '\\');
+			if (PHP_VERSION_ID >= 70000 && $reflection->hasReturnType()) {
+				$type = $reflection->getReturnType();
+				$class = !$type->isBuiltIn() ? (string) $type : null;
```
dg: Why is type null when is built-in?

src/DI/ContainerBuilder.php
dg: If it doesn't break anythink, it is not BC break.

src/DI/ContainerBuilder.php
dg: Ok, order is not important here.

src/DI/ContainerBuilder.php
pavelkouril: I hope I didn't misunderstood the code (and it's intent and use cases), but it seems it's just for resolving classes, not every type.

If you add support for resolving builtin type and then have a factory with `createString(): string`, when using it (ie. `@x::createString`), resolving fails with `Nette\DI\ServiceCreationException: Class or interface string used in service 'y' not found.` - and since scalar type can't be a service (or can it?), I find resolving it pointless. Maybe an exception that you can't resolve builtin types should be thrown, though?

dg: Exactly, it should throw exception. Maybe `Class or interface string` can be changed to `Type string`, but it should throw it.

pavelkouril: So you'd let it resolve builtin type correctly and then let `resolveServiceClass()` fail?

Isn't making the `resolveEntityClass()` itself throw exception (ie. `Nette\DI\ServiceCreationException: Type 'string' can not be used as return type in factory 'A\Factory::createString'`) a better solution?

dg: I do not know, but `@return string` and `class: string` (in config.neon) should fail too and solutions should be as easy as possible.

pavelkouril: Well, `class: string` fails anyways on [other exception](https://github.com/nette/di/blob/master/src/DI/ContainerBuilder.php#L481).

Also, I found out about some problematic cases with the `@return` expanding (like, `@return string` in namespace becomes `Foo\string` - not a problem in 5.x, but in 7.0), so the exception in `resolveEntityClass()` is probably a better solution.

dg: Maybe it can be fixed in expandClassName, ie. to not expand built-in types.

pavelkouril: Hehe, why can't I come up with these simple solutions. Thanks for the idea! It will be probably better than the solution I pushed, so I'll revert it and change `expandClassName`.

pavelkouril:
Is there something blocking this or can I rebase it and squash it?

tests/DI/Compiler.generatedFactory.returnTypes.phpt
```diff
@@ -0,0 +1,43 @@
+<?php
+
+/**
+ * Test: Nette\DI\Compiler: generated services factories from interfaces with return type declarations.
+ * @phpVersion 7.0
+ */
+
+use Nette\DI;
+use Tester\Assert;
+use Nette\DI\Config\Adapters as Adapt;
```
Majkl578: ?

pavelkouril: A little leftover from when I used another test as a template for this one. Removed. :)

src/DI/PhpReflection.php
```diff
@@ -81,6 +87,9 @@ public static function expandClassName($name, \ReflectionClass $rc)
 		if (empty($name)) {
 			throw new Nette\InvalidArgumentException('Class name must not be empty.');
 
+		} elseif (in_array($name, PHP_VERSION_ID >= 70000 ? ['string', 'int', 'float', 'bool', 'array', 'callable'] : ['array', 'callable'], TRUE)) {
```
dg: I think that version checking is not required here, so `in_array($name, ['string', 'int', 'float', 'bool', 'array', 'callable'], TRUE)` is enough.

dg:
@pavelkouril now it looks really very cool. The last thing: what about to add new method `PhpReflection::getReturnType()`?

src/DI/PhpReflection.php
```diff
@@ -59,13 +59,19 @@ public static function getDeclaringClass(\ReflectionProperty $prop)
...
 	public static function getPropertyType(\ReflectionParameter $prop)
 	{
-		try {
-			return ($ref = $prop->getClass()) ? $ref->getName() : NULL;
-		} catch (\ReflectionException $e) {
-			if (preg_match('#Class (.+) does not exist#', $e->getMessage(), $m)) {
-				return $m[1];
+		if (PHP_VERSION_ID >= 70000) {
+			return (string) $prop->getType();
+		} elseif ($prop->isArray()) {
```
dg: here should be added support for isCallable() too.

src/DI/PhpReflection.php
pavelkouril: It is needed. Naming class `string` is valid in PHP < 7.0 and it needs to be expanded if it's namespaced. Should I add a test for this, so the check isn't removed accidentaly in the future?

Majkl578: Also it should be case-insensitive check ('self' below as well).

pavelkouril: Should I use `strtolower` to `Nette\Utils\Strings::lower`? Because the latter one makes tests fail unless mbstring extension is loaded. How do you add the requirement for the tests to use some extension (travis, running it manually, etc.)?

src/DI/PhpReflection.php
pavelkouril: Oh, I understood that you want it in a separate PR afterwards (based on the discussion in this thread on outdated diff), but I'll add it now then and we'll just rename the commit to something like "added support for PHP 7 type hints and PHP 5 callable"? I have it prepared in stash, so it doesn't matter to me if it goes in in this PR on next one. :)

src/DI/ContainerBuilder.php
```diff
@@ -355,12 +355,17 @@ private function resolveImplement(ServiceDefinition $def, $name)
 		$def->setImplementType($methodName = $rc->hasMethod('create') ? 'create' : 'get');
 
 		if (!$def->getClass() && !$def->getEntity()) {
-			$returnType = PhpReflection::parseAnnotation($method, 'return');
-			if (!$returnType) {
-				throw new ServiceCreationException("Method $interface::$methodName() used in service '$name' has no @return annotation.");
+			if (PHP_VERSION_ID >= 70000 && $method->hasReturnType()) {
+				$returnType = (string) $method->getReturnType();
+			} else {
```
fprochazka: This should really have a `@return annotation fallback`, simple `else` will cause problems. This would be a BC Break - an unnecesary one.

src/DI/ContainerBuilder.php
```diff
@@ -463,9 +468,13 @@ private function resolveEntityClass($entity, $recursive = [])
 				throw new ServiceCreationException(sprintf("Factory '%s' used in service '%s' is not callable.", Nette\Utils\Callback::toString($entity), $name[0]));
 			}
 
-			$class = preg_replace('#[|\s].*#', '', PhpReflection::parseAnnotation($reflection, 'return'));
-			if ($class) {
-				$class = $refClass ? PhpReflection::expandClassName($class, $refClass) : ltrim($class, '\\');
+			if (PHP_VERSION_ID >= 70000 && $reflection->hasReturnType()) {
+				$class = (string) $reflection->getReturnType();
+			} else {
```
fprochazka: also here - https://github.com/nette/di/pull/77/files#r38300671

src/DI/ContainerBuilder.php
enumag: @fprochazka Doesn't the ` && $method->hasReturnType()` part take care of that?

fprochazka: that's why I've added `<del>` to my comment :)

src/DI/PhpReflection.php
dg: `callable` is used here https://github.com/pavelkouril/di/commit/6649b249e4b4b3e812e1c5baef689f843e6596a0#diff-49f23e184ca8ecc097b83d102bfdc2a0R90, so PR is not needed, but ok, I'll prepare support for 2.3/master.

src/DI/PhpReflection.php
dg: Yes, `string` is valid class in PHP < 7, but it is very unlikely that `@return string` in PHP < 7 means class named `string`.

dg:
@pavelkouril I prepared support for native types https://github.com/nette/di/commit/e924c9835887b5bbaf192bbf91742164b50710ea & https://github.com/nette/di/commit/6a7306b00e737e83c1447eb037f6893dbbbee563 and added method getReturnType https://github.com/nette/di/commit/0a405d9799ef4e45e2f62bd0bc92e77bc393b5ce - can you rebase it?

pavelkouril:
Ok, I rebased and forcepushed it - now the modifications are really simple

dg:
I found some details: NULL and FALSE should be uppercase, one `use` keyword per declaration and space should be in `function ()`.

pavelkouril:
Forcepushed again; the `use` declaration coding style is inconsistent though, because some tests (for example `ContainerBuilder.factory.resolveClass.phpt` which I used as a basis for my tests) uses the older CS.

I'm willing to prepare PR to fix that, but I'm leaving for a week in few hours, so I can only do it when I'm back. :)

dg:
Thank you!

fprochazka:
@pavelkouril very nice, good job!

vojtech-dobes:
This is great to see :). I admire work of all of you.

lookyman:
:+1:

PR #80, opened 2015-08-24
PHP 5.6: Use argument unpacking

Majkl578:
Depends on  #79 (composer.json & Travis).

### DIFF
--- a/src/DI/Container.php
+++ b/src/DI/Container.php
@@ -154,7 +154,7 @@ class Container extends Nette\Object
 
 		try {
 			$this->creating[$name] = TRUE;
-			$service = call_user_func_array([$this, $method], $args);
+			$service = $this->$method(...$args);
 
 		} finally {
 			unset($this->creating[$name]);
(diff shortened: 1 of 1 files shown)


src/DI/ContainerLoader.php
```diff
@@ -108,7 +108,7 @@ protected function generate($class, $generator)
 	{
 		$compiler = new Compiler;
 		$compiler->getContainerBuilder()->setClassName($class);
-		$code = call_user_func_array($generator, [& $compiler]);
+		$code = $generator(...[& $compiler]);
```
hrach: `$generator(& $compiler)` ?

JanTvrdik: removing `call_user_func_*` will not work in all cases, e.g. if `$generator` is string.

Majkl578: @hrach: That is a syntax error in PHP 7.
What format exactly? I think only static methods? That would be eventually solved when we switch to PHP 7.

hrach: Oh I see, this is a nasty hack to workaround call-time pass-by-reference. Imo we should get rid of it anyway :)

dg: Nice hack, but for PHP 5.6 is `call_user_func_array` required here.

Majkl578: Ok, reverted & rebased.

PR #81, opened 2015-08-28
Fixes because of PHP7 strict mode

kukulich:


src/DI/Config/Adapters/NeonAdapter.php
```diff
@@ -37,20 +37,21 @@ private function process(array $arr)
...
-				$val[Helpers::EXTENDS_KEY] = Helpers::OVERWRITE;
-
-			} elseif (preg_match('#^(\S+)\s+' . self::INHERITING_SEPARATOR . '\s+(\S+)\z#', $key, $matches)) {
-				if (!is_array($val) && $val !== NULL) {
-					throw new Nette\InvalidStateException("Inheritance operator is available only for arrays, item '$key' is not array.");
-				}
-				list(, $key, $val[Helpers::EXTENDS_KEY]) = $matches;
-				if (isset($res[$key])) {
-					throw new Nette\InvalidStateException("Duplicated key '$key'.");
+			if (is_string($key)) {
```
dg: Please can you change the deep of nesting, ie. use `is_string($key)` in both conditions or `if (!is_string($key)))` as first condition?

kukulich:
Forcepushed.

src/DI/ContainerBuilder.php
```diff
@@ -266,7 +266,9 @@ public function autowireArguments($class, $method, array $arguments)
 		if (!$rm->isPublic()) {
 			throw new ServiceCreationException("$class::$method() is not callable.");
 		}
-		$this->addDependency($rm->getFileName());
+		if (!$rm->isInternal()) {
```
dg: `addDependency` is used 3× times (here too https://github.com/nette/di/blob/ea244d0662c164a164bb91e622b00dc361f0668c/src/DI/ContainerBuilder.php#L696), so what about `$this->addDependency((string) $rm->getFileName())` and keep it simple?

src/DI/ServiceDefinition.php
```diff
@@ -48,7 +48,7 @@ class ServiceDefinition extends Nette\Object
 	 */
 	public function setClass($class, array $args = [])
 	{
-		$this->class = ltrim($class, '\\');
+		$this->class = ltrim((string) $class, '\\');
```
dg: `$this->class = $class ? ltrim($class, '\\') : NULL;` seems better here

kukulich:
Force pushed again.

dg:
Great, thanks!

src/DI/ContainerBuilder.php
```diff
@@ -335,7 +335,8 @@ public function prepareClassList()
 		}
 
 		foreach ($this->classes as $class => $foo) {
-			$this->addDependency((new ReflectionClass($class))->getFileName());
+			$rc = new ReflectionClass($class);
```
dg: Ups, there should remain `(new ReflectionClass($class)`… I'll reforce push it.

kukulich: Ok, it was more readable for me so I let it that way.

PR #83, opened 2015-09-06
Fixed #82 (autowiring of builtin types with default value)

### LINKED ISSUE #82: Broken creation of services that have optional array constructor dependency
if you have a class with constructor like this

    public function __construct(array $server = array(), History $history = null, CookieJar $cookieJar = null)
    {
        $this->setServerParameters($server);
        $this->history = $history ?: new History();
        $this->cookieJar = $cookieJar ?: new CookieJar();
    }

and config.neon 

    - Goutte\Client

this exception is raised exception 'Nette\DI\ServiceCreationException' with message 'Class array needed by Symfony\Component\BrowserKit\Client::__construct() not found. Check type hint and 'use' statements.'

this class is just an example, it happens with all array typehints. it started to happen in recent versions of nette

before it worked just fine

pavelkouril:
Fix for #82 - problem was introduced by commit https://github.com/nette/di/commit/6a7306b00e737e83c1447eb037f6893dbbbee563

I hope this is the correct fix :)

### DIFF
--- a/src/DI/Helpers.php
+++ b/src/DI/Helpers.php
@@ -88,6 +88,7 @@ class Helpers
 			. $method->getName() . '()';
 
 		foreach ($method->getParameters() as $num => $parameter) {
+			$class = PhpReflection::getParameterType($parameter);
 			if (array_key_exists($num, $arguments)) {
 				$res[$num] = $arguments[$num];
 				unset($arguments[$num]);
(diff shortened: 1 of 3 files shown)


src/DI/Helpers.php
```diff
@@ -98,7 +99,7 @@ public static function autowireArguments(\ReflectionFunctionAbstract $method, ar
 				unset($arguments[$parameter->getName()]);
 				$optCount = 0;
 
-			} elseif ($class = PhpReflection::getParameterType($parameter)) { // has object type hint
+			} elseif ($class !== NULL && !in_array(strtolower($class), ['string', 'int', 'float', 'bool', 'array', 'callable'], TRUE)) { // has object type hint
```
fprochazka: this would be nicer as some `PhpReflection::isNativeType()` function

pavelkouril: Yeah, I was thinking the same (probably `isBuiltinType` to be consistent with PHP 7 naming), since the same `in_array` is used few lines lower in `expandClassName`. But I'm pretty new to Nette PRs and I'm not sure if such refactoring is acceptable in PR with a bugfix?

fprochazka: @pavelkouril if you do it in two commits, then why not :)

dg:
Thanks, merged (in the opposite order of commits).

PR #84, opened 2015-09-10
Alternative autowiring approach [WIP]

dg:
~~All services are autowired. `setAutowired(TRUE)` means that service is preferred, `setAutowired(FALSE)` means that is autowired too, but not preferred (yes, name `setAutowired(FALSE)` is weird here, but this is just experiment).~~

Only services that are excluded via `$excludedClasses` are not autowired.

~~Now you can create services FileStorage (autowired: on) and DevNullStorage (autowired: off) and access the first one via `getByType('IStorage') and second one via `getByType('DevNullStorage')`.~~

I tried [a different approach](https://github.com/nette/di/pull/84#issuecomment-217701242)

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -349,11 +349,38 @@ class ContainerBuilder
 		}
 
 		//  build auto-wiring list
-		$this->classList = [];
+		$this->classList = $preferred = [];
 		foreach ($this->definitions as $name => $def) {
 			if ($class = $def->getImplement() ?: $def->getClass()) {
+				$defAutowired = $def->getAutowired();
+				if (is_array($defAutowired)) {
+					foreach ($defAutowired as $k => $aclass) {
+						if ($aclass === self::THIS_SERVICE) {
+							$defAutowired[$k] = $class;
+						} elseif (!is_a($class, $aclass, TRUE)) {
+							throw new ServiceCreationException("Incompatible class $aclass in autowiring definition of service '$name'.");
+						}
+					}
+				}
+
 				foreach (class_parents($class) + class_implements($class) + [$class] as $parent) {
-					$this->classList[$parent][$def->isAutowired() && empty($this->excludedClasses[$parent])][] = (string) $name;
+					$autowired = $defAutowired && empty($this->excludedClasses[$parent]);
+					if ($autowired && is_array($defAutowired)) {
+						$autowired = FALSE;
+						foreach ($defAutowired as $aclass) {
+							if (is_a($parent, $aclass, TRUE)) {
+								if (empty($preferred[$parent]) && isset($this->classList[$parent][TRUE])) {
+									$this->classList[$parent][FALSE] = array_merge(...$this->classList[$parent]);
+									$this->classList[$parent][TRUE] = [];
+								}
+								$preferred[$parent] = $autowired = TRUE;
+								break;
+							}
+						}
+					} elseif (isset($preferred[$parent])) {
+						$autowired = FALSE;
+					}
+					$this->classList[$parent][$autowired][] = (string) $name;
 				}
 			}
 		}
@@ -455,7 +482,7 @@ class ContainerBuilder
 			}
 			self::checkCase($class);
 
-		} elseif ($def->isAutowired()) {
+		} elseif ($def->getAutowired()) {
 			trigger_error("Type of service '$name' is unknown.", E_USER_NOTICE);
 		}
 		return $class;
--- a/src/DI/ServiceDefinition.php
+++ b/src/DI/ServiceDefinition.php
@@ -36,7 +36,7 @@ class ServiceDefinition
 	/** @var array */
 	private $tags = [];
 
-	/** @var bool */
+	/** @var bool|string[] */
 	private $autowired = TRUE;
 
 	/** @var bool */
@@ -210,21 +210,30 @@ class ServiceDefinition
 
 
 	/**
-	 * @param  bool
+	 * @param  bool|string|string[]
 	 * @return self
 	 */
 	public function setAutowired($state = TRUE)
 	{
 		call_user_func($this->notifier);
-		$this->autowired = (bool) $state;
+		$this->autowired = is_string($state) || is_array($state) ? (array) $state : (bool) $state;
 		return $this;
 	}
 
 
 	/**
-	 * @return bool
+	 * @return bool|string[]
 	 */
 	public function isAutowired()
+	{
+		return $this->autowired;
+	}
+
+
+	/**
+	 * @return bool|string[]
+	 */
+	public function getAutowired()
 	{
 		return $this->autowired;
 	}
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -396,7 +396,7 @@ class Compiler
 		}
 
 		if (isset($config['autowired'])) {
-			Validators::assertField($config, 'autowired', 'bool');
+			Validators::assertField($config, 'autowired', 'bool|string|array');
 			$definition->setAutowired($config['autowired']);
 		}
 
--- a/tests/DI/ContainerBuilder.autowiring.types.phpt
+++ b/tests/DI/ContainerBuilder.autowiring.types.phpt
@@ -1,0 +1,201 @@
+<?php
+
+/**
+ * Test: Nette\DI\ContainerBuilder and direct types
+ */
+
+use Nette\DI;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+interface IFoo
+{
+}
+
+interface IBar
+{
+}
+
+class Foo implements IFoo
+{
+}
+
+class Bar extends Foo implements IBar
+{
+}
+
+
+test(function () {
+	$builder = new DI\ContainerBuilder;
+	$builder->addDefinition('bar')
+		->setClass('Bar')
+		->setAutowired('Bar');
+
+	Assert::same('bar', $builder->getByType('Bar'));
+	Assert::same(NULL, $builder->getByType('IBar'));
+	Assert::same(NULL, $builder->getByType('Foo'));
+	Assert::same(NULL, $builder->getByType('IFoo'));
+});
+
+
+test(function () {
+	$builder = new DI\ContainerBuilder;
+	$builder->addDefinition('bar')
+		->setClass('Bar')
+		->setAutowired('IBar');
+
+	Assert::same('bar', $builder->getByType('Bar'));
+	Assert::same('bar', $builder->getByType('IBar'));
+	Assert::same(NULL, $builder->getByType('Foo'));
+	Assert::same(NULL, $builder->getByType('IFoo'));
+});
+
+
+test(function () {
+	$builder = new DI\ContainerBuilder;
+	$builder->addDefinition('bar')
+		->setClass('Bar')
+		->setAutowired('Foo');
+
+	Assert::same('bar', $builder->getByType('Bar'));
+	Assert::same(NULL, $builder->getByType('IBar'));
+	Assert::same('bar', $builder->getByType('Foo'));
+	Assert::same(NULL, $builder->getByType('IFoo'));
+});
+
+
+test(function () {
+	$builder = new DI\ContainerBuilder;
+	$builder->addDefinition('bar')
+		->setClass('Bar')
+		->setAutowired('IFoo');
+
+	Assert::same('bar', $builder->getByType('Bar'));
+	Assert::same(NULL, $builder->getByType('IBar'));
+	Assert::same('bar', $builder->getByType('Foo'));
+	Assert::same('bar', $builder->getByType('IFoo'));
+});
+
+
+test(function () {
+	$builder = new DI\ContainerBuilder;
+	$builder->addDefinition('bar')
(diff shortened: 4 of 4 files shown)


dg:
cc @matej21

matej21:
:+1:

dg:
I wonder how this commit affects large containers (might it try @fprochazka, @ondrejmirtes, @pilec, …)? In my projects containers are the same, only with minor difference in meta data.

ondrejmirtes:
I use setting autowired to off to exclude services from being accidentally injected somewhere where I do not want them. For example I have all instances of Monolog\Logger set to autowired: off because when I ask for a logger in a constructor, I want the DI container to fail and require specifying which concrete logger I want to inject into the service.

I am afraid that this change will lead to injecting of random unwanted services. Or is my scenario somehow solvable after this change?

src/DI/ContainerBuilder.php
```diff
@@ -327,7 +331,8 @@ public function prepareClassList()
 		foreach ($this->definitions as $name => $def) {
 			if ($class = $def->getImplement() ?: $def->getClass()) {
 				foreach (class_parents($class) + class_implements($class) + [$class] as $parent) {
-					$this->classes[$parent][$def->isAutowired() && empty($excludedClasses[$parent])][] = (string) $name;
+					$autowired = isset($excludedClasses[$parent]) ? NULL : $def->isAutowired();
```
matej21: This breaks ApplicationExtension which looks for presenters by excluded class name: https://api.nette.org/2.3.5/source-Bridges.ApplicationDI.ApplicationExtension.php.html#97

dg:
@ondrejmirtes this can be solved in two ways:

1) option autowired will have three states: `off`, `on` and `preferred` (the default one), so you can disable autowiring in the same way as before
2) option autowired will have two states: `on` and `preferred` (the default one?) and you can disable autowiring via [excluded classes](https://github.com/nette/di/blob/6b24a1a420b05dd058b5e75510c117e31dfa26b6/src/DI/ContainerBuilder.php#L494) and they will be configurable via config.neon.

I have absolutely no idea which approach is better.

The 1) is close to current approach, but three states will make it very difficult for programmers to figure out which state should be used. 

The 2) fits perfectly to your scenario. You can disable autowiring for loggers with single line, so it reduces the potential for error or mistake.

src/DI/ContainerBuilder.php
dg: findByType() is not affected.

matej21: Oh, I didn't notice commit 6b24a1a420 :)

fprochazka:
I've tried it and it doesn't exactly work :) But I don't have much time to dig in.

snimek obrazovky porizeny 2015-09-11 20 49 35

MartinMystikJonas:
@dg: Just idea: What abou setAutowired($priority). FALSE = not autowire, TRUE=1=autowired lowest priority, 2+=autowired with ascending priority.

dg:
@MartinMystikJonas I think it is the same as first option here https://github.com/nette/di/pull/84#issuecomment-139355221

MartinMystikJonas:
@dg Basically yes. But my solution can solve this for deeper inheritance hierarchies when you need to set prefferences when multiple services in inheritance hierarchies matches searched type. 

For example:
Service A - `IStorage, FileStorage` `->setAutowire(3)`
Service B - `IStorage, FileStorage, MyFileStorage` `->setAutowire(2)`
Service C - `IStorage, DevNullStorage` `->setAutowire(1)`

`getByType('IStorage')` -> Service A
`getByType('FileStorage')` -> Service A
`getByType('MyFileStorage')` -> Service B
`getByType('DevNullStorage')` -> Service C

dg:
Service A with setAutowired(2) and B & C with setAutowired(1) will have the same result, or isn't it?

MartinMystikJonas:
You are right. My solution is only useful in even more complex situations and therefore probably not needed.

JanTvrdik:
:+1: Just found out I need this.

enumag:
One thing that I don't like on Nette/DI is overriding of autowired services by another extension. For example I've an extension which provides alternative implementation of Nette/Application/Application. Now what the extension can do to override the default implementation? In my opinion it's wrong for the custom extension to change the original service in any way (changing class/factory, turning of autowiring etc.) - but it currently can't be solved without it.

This PR doesn't solve this of course. I'm mentioning it because it could solve it (by using priorities or something). If you think it's off-topic, just ignore this comment.

brabijan:
:+1:

paveljanda:
:+1:

dg:
Now option `autowired` can contain name or names of classes/interfaces. It make services autowired only for specified types  and preferably.

src/DI/ContainerBuilder.php
```diff
@@ -348,11 +347,37 @@ public function prepareClassList()
...
 
 		//  build auto-wiring list
-		$this->classList = [];
+		$this->classList = $preferred = [];
 		foreach ($this->definitions as $name => $def) {
 			if ($class = $def->getImplement() ?: $def->getClass()) {
+				$defAutowired = $def->isAutowired();
+				if (is_array($defAutowired)) {
+					foreach ($defAutowired as $k => $aclass) {
+						if ($aclass === self::THIS_SERVICE) {
```
JanTvrdik: So you can write `autowire: [self]`? Is that useful?

dg: It is autowired only when typehint === class name.

src/DI/ContainerBuilder.php
```diff
@@ -348,11 +347,37 @@ public function prepareClassList()
...
+					$autowired = $defAutowired && empty($this->excludedClasses[$parent]);
+					if ($autowired && is_array($defAutowired)) {
+						$autowired = FALSE;
+						foreach ($defAutowired as $aclass) {
+							if (is_a($parent, $aclass, TRUE)) {
+								if (empty($preferred[$parent]) && isset($this->classList[$parent][TRUE])) {
+									$this->classList[$parent][FALSE] = array_merge(...$this->classList[$parent]);
+								}
+								$preferred[$parent] = $autowired = TRUE;
+								break;
```
JanTvrdik: Shouldn't we throw an exception if `!empty($preferred[$parent])`, or even better try to resolve who should win based on distance of `$parent` and `$aClass`? Now this seems to depend order of definitions.

src/DI/ContainerBuilder.php
JanTvrdik: Yes, but in what instances would you write `self` instead of writing directly the class name?
Also, it's not autowired when typehint === class name but when typehint instanceof class name, or not?

JanTvrdik:
Because you choose the *instance of approach* instead of *=== approach*, I don't understand why would you ever need to declare multiple autowire types, i.e. why support bool|array|string and not just bool|string

src/DI/ContainerBuilder.php
dg: It should not rely on order, but maybe is implementation wrong, I'll check it later.

dg:
I tried to implement a completely different solution. Option `autowired` can be true or false (the same meaning), or class name or list of class names.

In this case, autowiring is limited only to for these classes. For example, when class implements three interfaces and I want to autowired it only for one or two of them, I can specify these interfaces. When *argument typehint* is instance of *specified interface*, service will be used for autowiring. 

When I use specify it's class name (or `autowired: self`), service will be therefore autowired only when *argument typehint* === *specified interface*.

At the same time, in addition to limiting the class list, the service becomes for that type preferred. So if there are more autowired services that meet the specified interface or class name, this service wins.

src/DI/ContainerBuilder.php
JanTvrdik: I looked at it again and it seems to be fine, the conflict is resolved in runtime.

dg: I think that there is missing line, it should erase [TRUE] list…

JanTvrdik: You mean that the following should work?

~~~php
test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired(TRUE);
	$builder->addDefinition('foo')
		->setClass('Foo')
		->setAutowired('Foo');

	Assert::same('foo', $builder->getByType('Foo'));
	Assert::same('bar', $builder->getByType('Bar'));
});
~~~

JanTvrdik:
I think I finally understand it now, the getByType logic can be rewritten in pseudocode as

~~~
coalesce(
	definitions.filter(_.class instanceof type).filter(type instance of _.autowired),
	definitions.filter(_.class instanceof type).filter(_.autowired === true)
)
~~~

Making the following 8 tests cover the core logic.

~~~php
test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired('Bar');

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same(NULL, $builder->getByType('IBar'));
	Assert::same(NULL, $builder->getByType('Foo'));
	Assert::same(NULL, $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired('IBar');

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same('bar', $builder->getByType('IBar'));
	Assert::same(NULL, $builder->getByType('Foo'));
	Assert::same(NULL, $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired('Foo');

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same(NULL, $builder->getByType('IBar'));
	Assert::same('bar', $builder->getByType('Foo'));
	Assert::same(NULL, $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired('IFoo');

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same(NULL, $builder->getByType('IBar'));
	Assert::same('bar', $builder->getByType('Foo'));
	Assert::same('bar', $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired(['IFoo', 'IBar']);

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same('bar', $builder->getByType('IBar'));
	Assert::same('bar', $builder->getByType('Foo'));
	Assert::same('bar', $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired(['Foo', 'Bar']);

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same(NULL, $builder->getByType('IBar'));
	Assert::same('bar', $builder->getByType('Foo'));
	Assert::same(NULL, $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired(['Foo', 'IBar']);

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same('bar', $builder->getByType('IBar'));
	Assert::same('bar', $builder->getByType('Foo'));
	Assert::same(NULL, $builder->getByType('IFoo'));
});


test(function () {
	$builder = new DI\ContainerBuilder;
	$builder->addDefinition('bar')
		->setClass('Bar')
		->setAutowired(['IFoo', 'Bar']);

	Assert::same('bar', $builder->getByType('Bar'));
	Assert::same(NULL, $builder->getByType('IBar'));
	Assert::same('bar', $builder->getByType('Foo'));
	Assert::same('bar', $builder->getByType('IFoo'));
});
~~~

JanTvrdik:
Having a class hierarchy graph (including interfaces of course), the `setClass()` marks one node (green) in graph and `setAutowired()` marks the second node (orange). The resulting autowired types are all nodes along the path from the `setClass()` node to the `setAutowired()` node.

untitled diagram 1

~~~php
test(function () {
    $builder = new DI\ContainerBuilder;
    $builder->addDefinition('bar')
        ->setClass('Bar')
        ->setAutowired('IFoo');

    Assert::same('bar', $builder->getByType('Bar'));
    Assert::same(NULL, $builder->getByType('IBar'));
    Assert::same('bar', $builder->getByType('Foo'));
    Assert::same('bar', $builder->getByType('IFoo'));
});
~~~

src/DI/ServiceDefinition.php
```diff
@@ -210,19 +210,19 @@ public function getTag($tag)
...
+		$this->autowired = is_string($state) || is_array($state) ? (array) $state : (bool) $state;
 		return $this;
 	}
 
 
 	/**
-	 * @return bool
+	 * @return bool|string[]
 	 */
 	public function isAutowired()
```
TomasVotruba: Method with `isX` name, that returns `bool` and `string` looks really confusing to me.

I'd keep preferred autowired types and TRUE/FALSE state separated.

dg:
@JanTvrdik I'll use your more explicit tests

PR #85, opened 2015-09-13
Compiler::parseServices: allow prefixing service name in the config

matej21:
ref nette/web-content#377
cc @Majkl578 @greeny

### DIFF
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -255,6 +255,9 @@ class Compiler
 				$builder->removeDefinition($name);
 				continue;
 			}
+			if ($namespace) {
+				$def = Helpers::prefixServiceName($def, $namespace);
+			}
 
 			$params = $builder->parameters;
 			if (is_array($def) && isset($def['parameters'])) {
--- a/src/DI/Helpers.php
+++ b/src/DI/Helpers.php
@@ -165,4 +165,29 @@ class Helpers
 		return $args;
 	}
 
+
+	/**
+	 * Prepend extension name to identifier or service name.
+	 * @param mixed
+	 * @param string
+	 * @return mixed
+	 */
+	public static function prefixServiceName($config, $namespace)
+	{
+		if (is_string($config)) {
+			if (strpos($config, '@extension.') === 0) {
+				$config = '@' . $namespace . '.' . substr($config, 11);
+			}
+		} elseif ($config instanceof Statement) {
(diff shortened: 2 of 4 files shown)


greeny:
Nice one! :+1:

fprochazka:
Nice!

dg:
What about something like

```
services:
	articles: MyBlog\ArticlesModel(@connection)
	comments: MyBlog\CommentsModel(@connection, @self.articles)
```

matej21:
@dg maybe it looks a bit better, but [this](https://github.com/nette/di/pull/85/files#diff-faebc8d5248ff70a5817645c8b8be6ffR10) won't work. But it is probably useless anyway.

dg:
Hmmm, passing service names to methods seems like antipattern.

matej21:
updated

dg:
Great!

I'm just not quite sure that `self` is the best name for this prefix.

matej21:
I'm not sure either since it is used for self-referencing service.

btw, maybe this could be moved to the ContainerBuilder so you can use `@self.foo` syntax in the CompilerExtension as well (instead of `$this->prefix('@foo')`)

greeny:
> btw, maybe this could be moved to the ContainerBuilder so you can use @self.foo syntax in the CompilerExtension as well (instead of $this->prefix('@foo'))

Good idea :+1:

What about `@this`, `@current` or `@scope`?

dg:
> btw, maybe this could be moved to the ContainerBuilder so you can use @self.foo syntax in the CompilerExtension as well (instead of $this->prefix('@foo'))

I can imagine that it could cause complications, for example when somebody will `addSetup('@self ...')` to service from a different scope.

dg:
`@current`, `@this` and `@self` sound similary. `@scope` sound fine, but has a different meaning in DI http://www.javaranch.com/journal/2008/10/dependency-injection-what-is-scope.html.

src/DI/Compiler.php
```diff
@@ -250,6 +250,9 @@ public static function parseServices(ContainerBuilder $builder, array $config, $
 			} elseif ($namespace) {
 				$name = $namespace . '.' . $name;
 			}
+			if ($namespace) {
```
dg: Can you move it bellow next condition?

matej21: done

dg:
Thanks!

matej21:
great! now I can finally fix the doc :)

PR #91, opened 2015-12-24
ContainerBuilder::removeAlias() removes aliases

mrtnzlml:
This feature request is similar to the function `removeDefinition` in `ContainerBuilder`. It can be useful to remove aliases as well. For example I am trying to do hot-reload already registered extensions inside of another extensions to add additional configuration. Unfortunately it's not possible without removing aliases (or maybe I don't know how). See [this use case](https://gist.github.com/mrtnzlml/21690953afb9dfad1824#file-compilerextension-php-L47).

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -149,6 +149,16 @@ class ContainerBuilder extends Nette\Object
 
 
 	/**
+	 * Removes the specified alias.
+	 * @return void
+	 */
+	public function removeAlias($alias)
+	{
+		unset($this->aliases[$alias]);
+	}
+
+
+	/**
 	 * Gets all service aliases.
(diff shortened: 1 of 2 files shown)


src/DI/ContainerBuilder.php
```diff
@@ -149,6 +149,16 @@ public function addAlias($alias, $service)
 
 
 	/**
+	 * Removes the specified alias.
+	 * @param $alias
```
dg: Can you change it to `@return void`?

mrtnzlml: Yes, I am sorry. I did not realize this comment doesn't make any sense. Changed and squashed...

dg: Thanks

PR #92, opened 2016-01-14
Container: added getServiceType()

lookyman:
Find out service class without the need to call createService().

All tests are passing, but there's a PHP startup error for same reason.

### DIFF
--- a/src/DI/Container.php
+++ b/src/DI/Container.php
@@ -107,6 +107,26 @@ class Container extends Nette\Object
 
 
 	/**
+	 * Gets the service type by name.
+	 * @param  string
+	 * @return string
+	 * @throws MissingServiceException
+	 */
+	public function getServiceType($name)
+	{
+		if (isset($this->meta[self::ALIASES][$name])) {
+			return $this->getServiceType($this->meta[self::ALIASES][$name]);
+
+		} elseif (isset($this->meta[self::SERVICES][$name])) {
+			return $this->meta[self::SERVICES][$name];
+
+		} else {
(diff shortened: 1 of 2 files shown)


src/DI/Container.php
```diff
@@ -107,6 +107,26 @@ public function getService($name)
 
 
 	/**
+	 * Gets the service class by name.
+	 * @param  string
+	 * @return object
+	 * @throws MissingServiceException
+	 */
+	public function getServiceClass($name)
```
fprochazka: This implies it can get the class for real, which means the code that resolves the class name in prepareClassList() should be moved to (ideally) some public static function and reused.

dg: @fprochazka  This is Container, not ContainerBuilder.

dg:
@lookyman there are errors because you are using `array()` instead of `[]`.

lookyman:
@dg Ahhh, thank you!

src/DI/Container.php
```diff
@@ -107,6 +107,26 @@ public function getService($name)
 
 
 	/**
+	 * Gets the service class by name.
+	 * @param  string
+	 * @return object
```
dg: `@return string`?

dg:
To consider: what about `Type` instead of `Class`, because there are methods `getByType` and `findByType`. This method can return interface too.

lookyman:
Sure, that is a good idea actually!

src/DI/Container.php
fprochazka: Oh, my bad .)

dg:
Thanks

PR #93, opened 2016-01-27
ContainerBuilder: removed support for ? (BC break)

dg:
We should Identify all cases when `?` is used, resolve them and remove a question mark (it is like eval).

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -557,10 +557,7 @@ class ContainerBuilder
 		$entity = $this->normalizeEntity($statement->getEntity());
 		$arguments = $statement->arguments;
 
-		if (is_string($entity) && Strings::contains($entity, '?')) { // PHP literal
-			trigger_error("Wildcard '?' used in '$entity' is deprecated.", E_USER_DEPRECATED);
-
-		} elseif ($service = $this->getServiceName($entity)) { // factory calling
+		if ($service = $this->getServiceName($entity)) { // factory calling
 			$params = [];
 			foreach ($this->definitions[$service]->parameters as $k => $v) {
 				$params[] = preg_replace('#\w+\z#', '\$$0', (is_int($k) ? $v : $k)) . (is_int($k) ? '' : ' = ' . PhpHelpers::dump($v));
@@ -678,7 +675,7 @@ class ContainerBuilder
 	/** @internal */
 	public function normalizeEntity($entity)
 	{
-		if (is_string($entity) && Strings::contains($entity, '::') && !Strings::contains($entity, '?')) { // Class::method -> [Class, method]
+		if (is_string($entity) && Strings::contains($entity, '::')) { // Class::method -> [Class, method]
 			$entity = explode('::', $entity);
 		}
 
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -558,6 +558,7 @@ class ContainerBuilder
 		$arguments = $statement->arguments;
 
 		if (is_string($entity) && Strings::contains($entity, '?')) { // PHP literal
+			trigger_error("Wildcard '?' used in '$entity' is deprecated.", E_USER_DEPRECATED);
 
 		} elseif ($service = $this->getServiceName($entity)) { // factory calling
 			$params = [];
--- a/src/DI/PhpGenerator.php
+++ b/src/DI/PhpGenerator.php
@@ -13,7 +13,6 @@ use Nette;
 use Nette\PhpGenerator\Helpers as PhpHelpers;
 use Nette\PhpGenerator\PhpLiteral;
 use Nette\Utils\Reflection;
-use Nette\Utils\Strings;
 
 
 /**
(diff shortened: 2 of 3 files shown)


enumag:
So hypothetically... what should I use instead of

```php
->addSetup('$service->methodA(?)->methodB(?)', ['a', 'b']);
```

dg:
`->addSetup([new Statement('@self::methodA', ['a']), 'methodB'], ['b']);`

enumag:
Yep, that's what I was afraid of. :-1:

dg:
Why :-1:? Is it forbidden to create some syntactic sugar? Like `chain('@self::methodA', ['a'], 'methodB'], ['b'])`?

enumag:
First its a BC break that would force me to update quite a few of my extensions. Second the syntax using question marks is probably the best syntactic sugar you could come up with - the chain syntax from your last comment doesn't solve cases like `@a->b(@c->d(?))`. Also this would effect neon definitions as well right?

By the way you didn't mention why you want to remove the question marks in the first place. Right now I fail to see any benefit, only BC break and uglier syntax. Thats why :-1: for now but you might be able to convince me.

dg:
Because eval is evil. DI compiler is unable to understand what it means. It relies on internal implementation details, like name `$service` etc…

enumag:
`$service` can be replaced with `@self` right?

fprochazka:
@dg I'm strongly against removing the ability to push custom code to the service creation. It's very handy and allows for hacking the container :)

I've went through most of my code, and the main reason for using the `?` is appending to an array. So adding a syntax like this would be neccesary

```php
$def->addSetup('$array[]', ["value"])
# converts to
$service->array[] = "value";
```

Imho it makes sense first to introduce new syntaxes to solve the causes for using the `?` and remove/depricate it much longer after it's adopted.

dg:
@fprochazka see https://github.com/nette/di/pull/93#issue-129077391

fprochazka:
@dg "eval is evil" is a statement taugh to begginers because they can create a lot of damage by using it wrong and also create exploits. It's not evil, just dangerous.

fprochazka:
@dg I never figured the array in setup's statement, that's a nice one.

dg:
> new syntaxes to solve

I think it can be solved with current syntax (`::array_push(@self::$array, value)` in neon), but I am opended to enhance syntax.

fprochazka:
@dg I get your point, but I hope I'm not the only one who considers this very ugly :)

fprochazka:
My use-cases for custom code in service creation

- [Clock/DI/ClockExtension.php](https://github.com/Kdyby/Clock/blob/4e101a0725f6d226f9435fc0de09c4b23badcb33/src/Kdyby/Clock/DI/ClockExtension.php#L62) ... hope you won't
- [Console/DI/ConsoleExtension.php](https://github.com/Kdyby/Console/blob/ef9715211caef00c1dbfa61213fcaf3f435c63d2/src/Kdyby/Console/DI/ConsoleExtension.php#L187-L192) this could be replaced by array append and some static function
- [Console/DI/ConsoleExtension.php](https://github.com/Kdyby/Console/blob/ef9715211caef00c1dbfa61213fcaf3f435c63d2/src/Kdyby/Console/DI/ConsoleExtension.php#L198-L202) this by better compile-time analysis of the service, but mainly array append
- [Doctrine/DI/OrmExtension.php](https://github.com/Kdyby/Doctrine/blob/c85d1f11ab9ddafd7c634f95a0638d80a8523b68/src/Kdyby/Doctrine/DI/OrmExtension.php#L560) for reusing the variable later, this can be refactored easily
- [Doctrine/DI/OrmExtension.php](https://github.com/Kdyby/Doctrine/blob/c85d1f11ab9ddafd7c634f95a0638d80a8523b68/src/Kdyby/Doctrine/DI/OrmExtension.php#L569) chaining, can be refactored with https://github.com/nette/di/pull/93#issuecomment-175505537 
- [Facebook/DI/FacebookExtension.php](https://github.com/Kdyby/Facebook/blob/7530f665a8f036acc7a19b8ec0a92bf4cb497f49/src/Kdyby/Facebook/DI/FacebookExtension.php#L78) array prepend... could be solved by introducing a method I guess
- [Facebook/DI/FacebookExtension.php](https://github.com/Kdyby/Facebook/blob/7530f665a8f036acc7a19b8ec0a92bf4cb497f49/src/Kdyby/Facebook/DI/FacebookExtension.php#L112) array append
- [Google/DI/GoogleExtension.php](https://github.com/Kdyby/Google/blob/957410b6efbf39939b62df6b265559a525e51802/src/Kdyby/Google/DI/GoogleExtension.php#L130) there was a cyclic dependency in the original library
- [HtmlValidator/DI/ValidatorExtension.php](https://github.com/Kdyby/HtmlValidatorPanel/blob/51df775e068a0c378c2ee00a59093d10d5089e24/DI/ValidatorExtension.php#L38-L40) array append
- [Translation/DI/TranslationExtension.php](https://github.com/Kdyby/Translation/blob/3f54b00d09fb273643e3f2bd8ac96212f06794d2/src/Kdyby/Translation/DI/TranslationExtension.php#L279-L285) array append

fprochazka:
@dg please reconsider the BC Break effect of deprecating this on extensions. I'm :+1: on adding the new syntax for array append, but :-1: on removing the `?`.

JanTvrdik:
It may be interesting to use https://wiki.php.net/rfc/use_function#domain_specific_languages combined with https://wiki.php.net/rfc/group_use_declarations instead of parsing strings.

fprochazka:
@JanTvrdik nice idea

dg:
Perhaps the solution would be to introduce less magic syntax, for example

```
		setup:
			- '$service->onClick[] = ?'([@self, 'method'])  # old way
			- ?php('$service->onClick[] = ?', [@self, 'method'])  # new way
			- method('? + ?'(1, 2))  # old way
			- method(?php('? + ?', 1, 2)) # new way
```

Issue #60, opened 2015-02-26
Parameters auto-resolution for generated factories depends on variable name


janedbal:
Following example:

```yaml
services:
    - ISomeControlFactory
```

```php
interface ISomeControlFactory {
    /** @return SomeControl */
    public function create(SomeEntity $someEntity);
}

class SomeControl {
    public function __construct(SomeEntity $entity) {}
}

class SomeEntity {}
```

leads to exception
```
Service 'XX_ISomeControlFactory': Service of type SomeEntity needed by SomeControl::__construct() not found. Did you register it in configuration file?
```

just because the parameter variables have different names.

lookyman:
I believe this is intentional, the automatic argument resolution matches them by their name. It could probably be improved, but it's a freakin' magic as it is, and if I have to use the same argument names to actually be able to understand my own code, then I'd prefere to do that..

enumag:
It is intentional. Mainly because constructor can contain other arguments that are either autowired or configured manually in config.neon or extension. I belive that different behaviour would be even more magical and could have some unexpected results.

The exception is wierd though. I will test it because the exception message should be different.

lookyman:
When `SomeControl`'s constructor needs `SomeEntity`, it takes it from `ISomeControlFactory`'s `create` method's argument, when they match. When they don't, it tries to take it from `Container` (and ignores the one from the argument). That is why the exception message is what it is.

enumag:
Yes I understand that. But I'd expect an exception to say somthing like 'Parameter '$someEntity' not found in 'SomeControl::__construct().'

enumag:
@janedbal By the way the service doesn't have to be anonymous.

castamir:
It leeds to "needle & haystack" problem... simply imagine 2 required parameters  of the same type (class) in `create()` method.

On the other hand, exception message should be better

enumag:
I tried to implement better exception but it broke [this test](https://github.com/nette/di/blob/master/tests/DI/ContainerBuilder.factory.params.phpt#L21-L24). It means that the arguments from create method can be used in setup, not necessarily in constructor. Which means the exception can't be improved.

PR #99, opened 2016-03-01
Check for unused parameter when generating factory

EdaCZ:
Purpose: throw an Exception when you have typo in the name of parameter in factory method (or constructor of factory product).

Related: #60

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -416,6 +416,9 @@ class ContainerBuilder extends Nette\Object
 						throw new ServiceCreationException("Type hint for \${$param->getName()} in $interface::$methodName() doesn't match type hint in $class constructor.");
 					}
 					$def->getFactory()->arguments[$arg->getPosition()] = self::literal('$' . $arg->getName());
+				} elseif (!$def->getSetup()) {
+					$hint = Nette\Utils\ObjectMixin::getSuggestion(array_keys($ctorParams), $param->getName());
+					throw new ServiceCreationException("Unused parameter \${$param->getName()} when implementing method $interface::$methodName()" . ($hint ? ", did you mean \${$hint}?" : '.'));
 				}
 				$paramDef = $hint . ' ' . $param->getName();
 				if ($param->isOptional()) {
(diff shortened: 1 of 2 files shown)


src/DI/ContainerBuilder.php
```diff
@@ -416,6 +416,8 @@ private function resolveImplement(ServiceDefinition $def, $name)
 						throw new ServiceCreationException("Type hint for \${$param->getName()} in $interface::$methodName() doesn't match type hint in $class constructor.");
 					}
 					$def->getFactory()->arguments[$arg->getPosition()] = self::literal('$' . $arg->getName());
+				} elseif (!$def->getSetup()) {
```
fprochazka: I'm confused, how is `setup` of service relevant to it's constructor parameters and to parameters of `create` method of the factory interface?

EdaCZ: When implementing generated factory there are only two ways to use parameters in `create` method:
a) in constructor of created object.
b) in setup directives.
Or am I wrong?

enumag: @fprochazka When the parameter from create method did not exist in the constructor it failed with no clear exception. I was trying to solve this when #60 was reported but my solution broke some tests - the parameter can be used in setup instead of the constructor. With this condition it should work fine.

enumag:
Good job @EdaCZ. :+1:

src/DI/ContainerBuilder.php
dg: Exception message should be even better :) IMHO typo is in interface, not in constructor, because constructor is primary and interface secondary.

xificurk:
Improvement suggestion: Add "Did you mean ...?" feature to the exception message.

src/DI/ContainerBuilder.php
EdaCZ: I will try to implement "Did you mean" feature.
This would be ok?
`Unused parameter $baz when implementing method Bad4::create(). Typo in Bad4::create() parameters? Did you mean $bar?`

f3l1x: What about: `Unused parameter $baz when implementing method Bad4::create(). Did you mean $bar in Bad4?`

EdaCZ: I am not sure, what to choose.
So, @dg, please choose one and I'll update commit.

EdaCZ:
@dg I've updated commit with "Did you mean feature". Is it ok now?

dg:
Great! Thanks

PR #100, opened 2016-03-06
readme.md: typo

grogy:


grogy:
Is something bad with this (small) PR? I mean that this construct lead developers for better code (great for refactoring, for reading and code consistant). And minimal version PHP for this library is is 5.6 - so this language construct is available.

dg:
Can you chante commit message to something like `readme.md: typo`?

grogy:
Of course, done.

dg:
Thanks

PR #102, opened 2016-03-10
Add support for @return $this in factories [closes #101]

### LINKED ISSUE #101: Return annotation containing $this confuses ContainerBuilder
I have a project that is using `@return $this` annotations instead of `@return self`, and such classes cannot be registered as services. The class resolving fails with following:
> Nette\DI\ServiceCreationException
> Type Mautic\Auth\$this used in service 'mauticApi' not found or is not class or interface.

You can see the file in question [directly here](https://github.com/mautic/api-library/blob/master/lib/Auth/ApiAuth.php#L17-L32). I think `return $this` is not according to the phpDoc spec, but nevertheless I think ContainerBuilder should support it.

Aurielle:
My take at fixing #101. Also included tests for this, I hope they're at least somewhat satisfactory.

### DIFF
--- a/src/DI/PhpReflection.php
+++ b/src/DI/PhpReflection.php
@@ -118,7 +118,7 @@ class PhpReflection
 		} elseif (self::isBuiltinType($lower)) {
 			return $lower;
 
-		} elseif ($lower === 'self' || $lower === 'static') {
+		} elseif ($lower === 'self' || $lower === 'static' || $lower === '$this') {
 			return $rc->getName();
 
 		} elseif ($name[0] === '\\') { // fully qualified name
--- a/tests/DI/ContainerBuilder.factory.resolveClass.phpt
+++ b/tests/DI/ContainerBuilder.factory.resolveClass.phpt
@@ -40,6 +40,27 @@ namespace B
 
 }
 
+namespace C
+{
+	class SelfFactory
+	{
+		/** @return self */
+		static function create()
+		{
+			return new self;
+		}
+	}
+
+	class ThisFactory
+	{
+		/** @return $this */
+		static function create()
+		{
+			return new self;
+		}
+	}
+}
+
 namespace
 {
 	use Nette\DI;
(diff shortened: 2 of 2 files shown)


src/DI/PhpReflection.php
```diff
@@ -118,7 +118,7 @@ public static function expandClassName($name, \ReflectionClass $rc)
 		} elseif (self::isBuiltinType($lower)) {
 			return $lower;
 
-		} elseif ($lower === 'self' || $lower === 'static') {
+		} elseif ($lower === 'self' || $lower === 'static' || $lower === '$this') {
```
fprochazka: :+1:

dg:
`$this` is not type…

What will you use here?

```php
/** @return ? */
function foo()
{
    return new static;
}
```

or here


```php
/** @return ? */
function foo()
{
    return clone $this;
}
```

src/DI/PhpReflection.php
Majkl578: :-1:

Aurielle:
@dg I'd use self or class name in both cases (maybe static in the first one, but prior to today I had no idea it even existed). I'm just hinting that some projects use $this and it'd be awesome if nette/di supported it (see the original issue please), instead of failing with "No such class" exception.

I think we all agree that this isn't an ideal solution, because (afaik) `@return $this` is not part of phpDoc spec. But since some projects use it and some IDEs support it (PhpStorm does), and maintaining forks of either the projects in question or nette/di is tiresome and not something everyone would like to do, I'd like to see this PR accepted so the libraries using this syntax could be registered as services as well. (I have written a different factory that uses correct typehint as a workaround in the meantime, but this is essentially code duplication.)

dg:
So please add containerbuilder to commit message:

Aurielle:
Commit message updated.

src/DI/PhpReflection.php
fprochazka: @Majkl578 phpstorm generates this stupid thing, do you know how to change it?

PR #104, opened 2016-03-26
Add support for @var statement[|statement...]

Trejjam:
Add possibility use IDE code completion based on PhpDoc

```php
class FooPresenter extends UI\Presenter {
  /** @var Nette\Mail\IMailer|Nette\Mail\SmtpMailer @inject */
  public $mailer;
}
```

### DIFF
--- a/src/DI/Extensions/InjectExtension.php
+++ b/src/DI/Extensions/InjectExtension.php
@@ -83,6 +83,10 @@ class InjectExtension extends DI\CompilerExtension
 			$rp = new \ReflectionProperty($class, $name);
 			if (PhpReflection::parseAnnotation($rp, 'inject') !== NULL) {
 				if ($type = PhpReflection::parseAnnotation($rp, 'var')) {
+					if (Nette\Utils\Strings::contains($type, '|')) {
+						list($type) = explode('|', $type); //Use only first statement
+					}
+
 					$type = PhpReflection::expandClassName($type, PhpReflection::getDeclaringClass($rp));
 				}
 				$res[$name] = $type;
--- a/tests/DI/InjectExtension.getInjectProperties().phpt
+++ b/tests/DI/InjectExtension.getInjectProperties().phpt
@@ -71,6 +71,28 @@ namespace C
 	}
 }
 
+namespace D
+{
+	interface DInterface
+	{
+
+	}
+
+	class DInjected implements DInterface
+	{
+
+	}
+
+	class DClass
+	{
+		/** @var DInterface|DInjected @inject */
+		public $var1;
+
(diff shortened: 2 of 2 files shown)


fprochazka:
This `Nette\Mail\IMailer|Nette\Mail\SmtpMailer` is a really bad practice. Either you depend on the interface, or you depend on the implementation. You do not depend on both.

Trejjam:
I know, this is bad example (there are no diference in method, ...).

I use it here:
```php
/** @var Nette\Application\IPresenterFactory|Nette\Application\PresenterFactory @inject */
public $presenterFactory;
```
I need `Nette\Application\PresenterFactory::unformatPresenterClass` which is not defined in `IPresenterFactory`

fprochazka:
IMHO The Clean Way™ is to [create an interface for that](https://github.com/nette/application/issues/101#issuecomment-148650056) and inject that interface, since that is really what you need, not a PresenterFactory.

enumag:
@Trejjam unformatPresenterClass is internal, you should not rely on it if possible.

Trejjam:
Yes, but may I implement implemented (&tested) code? :)
Dne 3. 4. 2016 10:21 napsal uživatel "Jáchym Toušek" <
notifications@github.com>:

Trejjam:
I understand that this solution is not too good. If I implement what @fprochazka [suggest](https://github.com/nette/di/pull/104#issuecomment-202032219), do you like it?

enumag:
@Trejjam Personally I do. @dg does not though.

JanTvrdik:
> unformatPresenterClass is internal

It should not be, there a lot of use cases for it.

dg:
@Trejjam I think that what Filip suggests is not solution for your case.

Trejjam:
It not solve PhpDoc OR, but it move required method under new interface. This interface could by required without OR (method(s) will be in interface).

dg:
Filip's solution is about collection of objects, but you want to inject one object.

Trejjam:
Sorry my fault, I talk about last part of suggestion.

I understand it in this way: Create interface `IPresenterMapping` and move `formatPresenterClass` and `unformatPresenterClass` to new class, that will be dependency for `PresenterFactory`.

So if you will need [completely ]different p-mapping you could easy change it independently on p-factory.

fprochazka:
You can have `IPresenterFormatter` implemented by the factory itself (or by class that the factory has as a dependency) and by adding not-autowired `IPresenterFormatter` objects you add more behaviour to it. So it will behave as a collection. This is a known pattern used [for example in doctrine](https://github.com/doctrine/dbal/tree/master/lib/Doctrine/DBAL/Logging).

PR #105, opened 2016-04-16
ContainerBuilder: better error message when class is not instantiable

matej21:
https://forum.nette.org/cs/22402-nejvetsi-zaseky-v-nette-drobnosti-na-kterych-se-da-viset-i-nekolik-hodin?p=2#p170795


dg:
Thanks!

PR #106, opened 2016-04-17
Fix BC break in Statement and ServiceDefinition

enumag:
```
E_USER_DEPRECATED: Missing annotation @property for Nette\DI\Statement::$entity
```

`$statement->entity` is commonly used by extensions so it's probably not a good idea to make this BC break before 3.0 (and even there it's questionable).

### DIFF
--- a/src/DI/ServiceDefinition.php
+++ b/src/DI/ServiceDefinition.php
@@ -12,6 +12,10 @@ use Nette;
 
 /**
  * Definition used by ContainerBuilder.
+ *
+ * @property $class
+ * @property $factory
+ * @property-read $entity
  */
 class ServiceDefinition
 {
--- a/src/DI/Statement.php
+++ b/src/DI/Statement.php
@@ -12,6 +12,8 @@ use Nette;
 
 /**
  * Assignment or calling statement.
+ *
+ * @property $entity
  */
 class Statement
 {


dg:
It must be declared on class.

src/DI/ServiceDefinition.php
```diff
@@ -12,6 +12,10 @@
 
 /**
  * Definition used by ContainerBuilder.
+ *
+ * @property $class
+ * @property $factory
+ * @property-read $entity
```
enumag: @dg Is this correct? Also there are other private properties with getters and setters. Should I add all of them? (These seem to be important.)

src/DI/Statement.php
```diff
@@ -12,6 +12,8 @@
 
 /**
  * Assignment or calling statement.
+ *
+ * @property $entity
  */
```
dg: there must be a type

enumag: @dg Like this? Does SmartObject support multiple types? (I didn't find such case it in nette/utils tests.)

```
@property string|array|ServiceDefinition|NULL $entity
```

dg: You should definitely test it before sending PR.

enumag: Yeah... it seemed like a small change at first. Sorry.

enumag:
I'm not quite sure which annotations should be added and which should not - there are many of them, some used commonly as properties other rarely.

Anyway I don't like to rely on magic properties anyway so I've fixed these occurences in Kdyby and my own packages instead.

The properties I had to fix were these:

- ServiceDefinition - factory, class, entity
- Statement - entity
- PhpGenerator\ClassType - methods

dg:
I think that commonly used properties should be added, because don't want to force users to change their code. But it is true that there are users which want to discover where magic properties are used and fix it.

So my plan is to release 2.4 beta soon, test it, decide which properties are commonly used and which are used rarely a then add properites to code.

PR #108, opened 2016-05-04
ContainerBuilder::getClassList: optimized performance

matej21:


src/DI/ContainerBuilder.php
```diff
@@ -259,14 +259,24 @@ public function findByTag($tag)
 	private function getClassList()
 	{
 		static $prev;
-		if ($this->classes !== FALSE && $prev !== serialize($this->definitions)) {
+		if ($this->classes !== FALSE && count($prev) !== count($this->definitions) && $prev !== $this->getDefinitionsInfo()) {
 			$this->prepareClassList();
-			$prev = serialize($this->definitions);
+			$prev = $this->getDefinitionsInfo();
```
JanTvrdik: You call `$this->getDefinitionsInfo` twice.

dg:
What's the trick?

matej21:
@dg the trick is that only class/factory/implement or autowire can affect the result of prepareClassList. e.g. setup or inject can't.

JanTvrdik:
What if you add one definitions and remove another one?

matej21:
@JanTvrdik good point. will fix..

matej21:
oh, it is a bit wrong. wait a minute :)

dg:
Where is the problem? Serialize takes a lot of time?

matej21:
In fact, that count is totally wrong. But without that performance goes from ~2s to ~8s (but still better than >30s)

@dg serialize is one problem. The second (and probably more important) is that prepareClassList is called many times, even it is not necessary (e.g. when setup method was added)

src/DI/ContainerBuilder.php
```diff
@@ -259,14 +259,24 @@ public function findByTag($tag)
 	private function getClassList()
 	{
 		static $prev;
-		if ($this->classes !== FALSE && $prev !== serialize($this->definitions)) {
+		if ($this->classes !== FALSE && count($prev) !== count($this->definitions) && $prev !== $this->getDefinitionsInfo()) {
```
castamir: You don't need to call `getDefinitionsInfo()` twice, do you?

JanTvrdik:
My guess is that you need to track dependencies of each individual class in class list.

castamir:
Got an idea, it's maybe a little offtopic, but still... is there a way, how to create a file watcher for e.g. PhpStorm? :D 

It might be interesting...

src/DI/ContainerBuilder.php
matej21: I do, its result is affected by prepareClassList

matej21:
rewritten

src/DI/ContainerBuilder.php
```diff
@@ -890,4 +898,13 @@ public function getServiceName($arg)
 		return $service;
 	}
 
+
+	/**
+	 * @internal
+	 */
+	public function notifyDefinitionChanged()
```
JanTvrdik: that's ugly why not just use events

matej21: I was worried about a performance. But it is ok... 

updated, what do you think now?

JanTvrdik:
But that's still slow, no?

src/DI/ContainerBuilder.php
```diff
@@ -59,14 +62,23 @@ class ContainerBuilder
...
 		if (isset($this->definitions[$name])) {
 			throw new Nette\InvalidStateException("Service '$name' has already been added.");
 		}
-		return $this->definitions[$name] = $definition ?: new ServiceDefinition;
+		if (!$definition) {
+			$definition = new ServiceDefinition();
+		}
+		$definition->onChange[] = function ($service, $method) {
+			if (in_array($method, ['setClass', 'setFactory', 'setImplement', 'setAutowired'], TRUE)) {
+				$this->definitionsChanged = TRUE;
```
enumag: Definitions changed even with the other methods. Maybe rename it to sth like classListObsolete?

matej21:
@JanTvrdik performance is great now. ~0.7s now vs ~0.5s without that commit which regenerates class list

src/DI/ContainerBuilder.php
matej21: Not sure I understand...

JanTvrdik:
Interesting. You still regenerate the class list more often that is necessary. But I guess that other cases are not nearly as important.

I was playing with [tracking dependencies per class](https://github.com/JanTvrdik/nette-di/commit/73a6ff34b44aaf4cedce4941ad5f30371b5b3ddc). But as it turned out, you can't really implement it this way without tracking calls to setImplement + setClass.

matej21:
> Interesting. You still regenerate the class list more often that is necessary. 

for my container it is regenerated ~10 times, that is ok

src/DI/ContainerBuilder.php
matej21: Oh, I understand now. Yes, that is better name :)

dg:
It would be great to put together a sandbox for doing DI benchmarks. With a greater number of services and with some extensions.

dg:
This is a nice solution, but unfortunately Statement is mutable too.

dg:
It may be finished it two ways:
1) replace __FUNCTION__ with attribute name to eliminate differences between setSetup() and addSetup()
2) make it as simple as possible and track only what we need, ie. setClass, setFactory, setImplement, setAutowired and additionally setEntity for entities added via setFactory. Notifier can be simple callback instead of array of callbacks.

I more like option 2.

matej21:
@dg how do you like the solution before this commit https://github.com/nette/di/pull/108/commits/9e680d94f62a5ffb17b006e0e0252646f6a821ef ? it was similar to your option 2

dg:
Yes, it is. It only differs in passed callback vs passed builder. I am not sure what is better.

dg:
Statement::setEntity is used extremely rare (two occurrences in Kdyby), probably best thing is to make it immutable in future and trigger warning in 2.4

matej21:
> I am not sure what is better.

Me neither. But passing CB is more straightforward and it cannot be abused by a user.

> probably best thing is to make it immutable in future and trigger warning in 2.4

:+1: so can I ignore it? and just revert commit  9e680d9?

dg:
Done 8211907ac01d5dc100fa01850c34d5a7873385e2

matej21:
PR updated. What do you think?

src/DI/ServiceDefinition.php
```diff
@@ -302,4 +316,14 @@ public function getInject()
 		return $this->getTag(Extensions\InjectExtension::TAG_INJECT);
 	}
 
+
+	/**
+	 * @param ContainerBuilder
+	 * @internal
+	 */
+	public function setContainerBuilder(ContainerBuilder $containerBuilder)
```
dg: Or optional constructor parameter?

dg:
I think that unserialize(serialize()) will create a clone of ContainerBuilder too.

dg:
Hey, I can add support for deep clone directly into ServiceDefinition, right?

matej21:
good point, updated. But do we still need a deep clone since Statement is (will be) immutable?

dg:
c0258ca0c204ca0d8d27d5ab66c64427d0263f8b

matej21:
rebased & squashed

dg:
What do you think about this https://github.com/dg/nette-di/commits/test - builder replaced with callback, so internal method is not needed

matej21:
@dg I'm ok with that.

dg:
$notifier = 'pi' is IMHO too tricky

matej21:
it is. maybe assign an empty closure in the constructor?

dg:
remove conditions + add constructor = 3 lines saved :-)

dg:
On the other hand, π trick is unique and Nette will be the first web framework that use this function…

PR #109, opened 2016-05-10
PhpGenerator

dg:


src/DI/ContainerBuilder.php
```diff
@@ -783,59 +582,64 @@ public function formatStatement(Statement $statement)
...
+		}
+	}
+
+
+	/**
+	 * Adds a file to the list of dependencies.
+	 * @return self
+	 * @internal
+	 */
+	public function addDependency($file)
```
fprochazka: This is actually a very usefull method and IMHO it shouldn't be marked internal.

src/DI/ContainerBuilder.php
```diff
@@ -174,24 +168,21 @@ public function getAliases()
 
 
 	/**
+	 * @param  string[]
 	 * @return self
 	 */
-	public function setClassName($name)
```
fprochazka: I see it's WIP, and you were probably gonna solve it anyway, but just so it doesn't disappear - removing this is a BC Break.

src/DI/ContainerBuilder.php
```diff
@@ -173,25 +173,6 @@ public function getAliases()
 	}
 
 
-	/**
-	 * @return self
-	 */
-	public function setClassName($name)
```
fprochazka: Do you intent to commit this to the `2.4.*` ?

src/DI/ContainerBuilder.php
dg: I know. I which case do you use it?

fprochazka: I never got to migrate and it's not really needed for me anymore (since you've solved it natively), but I was using custom container naming back in the day in tests, so it wouldn't collide.

fprochazka: Most likely nobody was using this method, but just removing seems a bit aggresive. But the trigger is good https://github.com/nette/di/commit/e3c82aa62d87506e33c8a73c2a87fecadf6b76fb :+1:

enumag: Actually I use it in some old tests. I have recently found a better way though.

dg:
Ready for merge

src/DI/Compiler.php
```diff
@@ -228,7 +228,8 @@ public function generateCode()
 			$this->dependencies[] = (new \ReflectionClass($extension))->getFileName();
 		}
 
-		$classes = $this->builder->generateClasses($this->className);
+		$generator = new PhpGenerator($this->builder);
+		$classes = $generator->generate($this->className);
```
JanTvrdik: What about passing the `$generator` to `afterCompile`?

JanTvrdik:
What's the recommended alternative to `formatPhp`? For example how to replace

~~~php
$logger = $builder->getDefinition($this->prefix('logger'));
$initialize->addBody($builder->formatPhp('Tracy\Debugger::setLogger(?);', [$logger]));
~~~

dg:
I do not know yet.

PR #112, opened 2016-05-22
Compiler: parseServices() and parseService() renamed to ??? [WIP]

dg:
It is named parseServices, because it parses 'service' in config, but it is public method and therefore should have more descriptive names. `addDefinitions()`? `loadDefinitions()`? `loadDefinitionsFromConfig()`?


f3l1x:
I like `loadDefinitions()`.

dg:
@f3l1x ok

PR #114, opened 2016-06-04
Compiler: add getExtension method

enumag:
In some cases (like Kdyby/Doctrine) there is a need for an extension to be configurable not only by config.neon but by other extensions as well. We have been using [this concept](https://forum.nette.org/en/18888-extending-extensions-solid-modular-concept) from @fprochazka for years now. The problem with this approach is that it creates a hard dependency - for example an extension which wants to configure `Kdyby\Doctrine\DI\OrmExtension` has to implement `Kdyby\Doctrine\DI\IEntityProvider`. That of course creates a hard depencency on Kdyby/Doctrine package which is not always desired.

The other (and in my opinion better) approach is to provide a public method which can be called by other extensions. There is only one small problem - while the other extension can access OrmExtension, it can't do so easily:

```php
    public function loadConfiguration()
    {
        $extensions = $this->compiler->getExtensions(OrmExtension::class);
        $extension = reset($extensions);
        // Some condition and exception if $extension is empty
        $extension->thePublicMethod($some, $arguments);
    }
```

Unfortunately this doesn't look good at all so I'd like to add one simple method to make this easier.

### DIFF
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -70,6 +70,25 @@ class Compiler
 
 
 	/**
+	 * @param string $class
+	 * @param bool $need
+	 * @return CompilerExtension
+	 */
+	public function getExtension($class, $need = true)
+	{
+		$extensions = $this->compiler->getExtensions($class);
+		$count = count($extensions);
+		if ($count > 1) {
+			throw new \Nette\Utils\AssertionException("Extension '$class' is installed $count times.");
+		}
+		if ($count < 1 && $need) {
+			throw new \Nette\Utils\AssertionException("Extension '$class' is not installed.");
+		}
+		return $count === 1 ? reset($extensions) : null;
+	}
+
+
+	/**
 	 * @return ContainerBuilder
 	 */
 	public function getContainerBuilder()
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -76,7 +76,7 @@ class Compiler
 	 */
 	public function getExtension($class, $need = true)
 	{
-		$extensions = $this->compiler->getExtensions($class);
+		$extensions = $this->getExtensions($class);
 		$count = count($extensions);
 		if ($count > 1) {
 			throw new \Nette\Utils\AssertionException("Extension '$class' is installed $count times.");


src/DI/Compiler.php
```diff
@@ -70,6 +70,25 @@ public function getExtensions($type = NULL)
...
+	 * @return CompilerExtension
+	 */
+	public function getExtension($class, $need = true)
+	{
+		$extensions = $this->getExtensions($class);
+		$count = count($extensions);
+		if ($count > 1) {
+			throw new \Nette\Utils\AssertionException("Extension '$class' is installed $count times.");
+		}
+		if ($count < 1 && $need) {
```
Majkl578: How about:
```php
if ($count === 0) {
    if ($need) {
        throw new \Nette\Utils\AssertionException("Extension '$class' is not installed.");
    }
    return NULL;
}
return reset($extensions);
```

Majkl578:
I'm not sure whether it's a good idea to throw an exception when multiple instances exist. Mostly because it's technically not forbidden/disallowed/discouraged concept (although quite rare imho).

src/DI/Compiler.php
```diff
@@ -70,6 +70,25 @@ public function getExtensions($type = NULL)
...
+	 * @param string $class
+	 * @param bool $need
+	 * @return CompilerExtension
+	 */
+	public function getExtension($class, $need = true)
+	{
+		$extensions = $this->getExtensions($class);
+		$count = count($extensions);
+		if ($count > 1) {
+			throw new \Nette\Utils\AssertionException("Extension '$class' is installed $count times.");
```
Majkl578: Using an exception from nette/utils doesn't seem right...

src/DI/Compiler.php
enumag: I prefer less nesting.

src/DI/Compiler.php
enumag: Suggest a better one. :-)

enumag:
@Majkl578 Yeah. But what else to do? Return the first one and ignore the others?

dg:
`$this->compiler->getExtensions(OrmExtension::class)[0]` seems fine to me.

enumag:
@dg The extensions are indexed by name (like `kdyby.doctrine`), not by numbers so `[0]` does not work.

dg:
Ok, so `current($this->compiler->getExtensions(OrmExtension::class))->…`. 

I'd like to solve these issues about ordering extensions etc. in next release and I feel that `getExtension()` is not something I'd like to support in future.

enumag:
@dg Understood. And thanks, current does make it simpler and returns false on empty array without causing an error (unlike Majkl's array_values + [0]).

Majkl578:
That's why I also added `?? NULL`.

enumag:
@Majkl578 I want to support PHP < 7 in my libraries.

Issue #116, opened 2016-06-14
Incorrectly instantiated DateTimes


repli2dev:
I attempted to use DateTime as parameter value in my config:
```
parameters:
    date: 2016-09-01
php:
    date.timezone: Europe/Prague
```

The parameter `date` is instanciated (that's cool), however this is done in the PHP default time zone which is (for example) `UTC` whereas I explicitly specify that I want to use `Europe/Prague`.

From brief look inside the source codes I understand that the sections are processed separetly, however this behaviour is highly unexpected and I would consider it a pitfall as it the bugs caused are not easy to find.

JanTvrdik:
Maybe PhpExtension can iterate over all parameters and set timezone to DateTime instances?

repli2dev:
Thanks for suggestion, I prepared a PR for that... Though the coercion of dates seems a bit tricky and from certain point of view the new behaviour could be a BC break as some people could depend on it. 
In that case I would suggest making this new behaviour optional.

dg:
Closed by https://github.com/nette/web-project/commit/a5e0f8c3394b22c0e554f6f38507a2fccbf8c7ac

PR #117, opened 2016-06-20
PhpExtension: fix incorrectly instantiated DateTimes

repli2dev:
Fixies issue #116

### DIFF
--- a/src/DI/Extensions/PhpExtension.php
+++ b/src/DI/Extensions/PhpExtension.php
@@ -8,6 +8,8 @@
 namespace Nette\DI\Extensions;
 
 use DateTime;
+use DateTimeImmutable;
+use DateTimeInterface;
 use DateTimeZone;
 use Nette;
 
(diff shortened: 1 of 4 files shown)


dg:
Isn't it too late to modify parameters in `afterCompile()`? I understand that this is not easy issue, it probably should be solve via some trick.

Majkl578:
Wouldn't this also modify parameters passed explicitly through Configurator::addParameters? That might be unexpected behavior.
And what about DateTimeImmutable?

repli2dev:
@dg I have moved it into `loadConfiguration` which seems also work. Also I have changed behaviour to support `DateTimeImmutable` too (along with DateTime for backward compatibility and cross version use).
Not sure what to do about Configurator::addParameters thought it seems to me as exactly the same problem we are trying to fix just in bootstrap or there is some case in which the fix is unwanted?

dg:
IMHO it will not work when you use parameter inside config, via %date%.

repli2dev:
Could you please elaborate? I am not sure that I understand your note about `%date%` -- is it relevant generally or to the Configurator::addParameters case? (Such as using only %date% in config an then providing this parameter later via the addParameter method?).

dg:
I mean something like 

```
parameters:
    date: 2016-09-01
services:
    - MyClass(%date%)
```

repli2dev:
I have added test showing that when using `loadConfiguration` then the provided use case works (though I have explicitly tested that it is not working with `afterCompile` nor `beforeCompile`).

I have also tested that using `Configurator::addParameters` in current way changes the timezone of all provided DateTimes which we don't know if it's desired. See some thoughts on the topic below.

------

If I am using `Configurator::addParameters` then I am in the bootstrap file before the container is configured, builded... executed. Therefore I am either using the timezone from PHP setting or my own (if I have manually added `date_default_timezone_set` there).

- **Case one**: manually added `date_default_timezone_set`: I know what I am doing, I do not want any of my datetimes to be touched!
- **Case two**: without anything, goes into two subcases
  - **Subcase A**: I have specified timezone of the created Datetime: I know what I am doing, I do not want any of my datetimes to be touched!
  - **Subcase B**: I have not specified timezone: I "expect" them to be in timezone I have configured.

Problem is that the expectation is probably different in config and in bootstrap file. Though I would choose some standardized approach.

Ideas:
 - Translation of all datetimes having the timezone same as the timezone from PHP configuration but that still ommits the case that I want some of my datetimes to be in that zone...
 - Translate only that datetimes coming from config files and ignore the others?

-----

Note: with closer look into allowed date formats accepted by the DateTime construtor I have found that there is even a way how to create dates with time offset (not timezone!) which could help but is not sufficient (as there is only the time offset not the timezone which matters).

dg:
Thanks for test, it really works, because PhpExtension is the first extension.
 
ad questions: what about a different approch? https://github.com/dg/sandbox/commit/7671402b4a85612ac8a911099dff583584feb807

PR #126, opened 2016-09-21
PhpReflection::getParameterType() can handle PHP 7.1 nullable types

MartinMajor:
Nullable parameter types are prefixed by `?` in Reflection in PHP 7.1 which causes that Nette autowiring doesn't work. Even though this behaviour has changed many times in PHP 7.1 ( https://github.com/php/php-src/commit/622d2f41d1cdb597f4fafecaaacf66e238742bd4 , https://github.com/php/php-src/commit/8855a2ce76e8bfba1d2eea1345c765fde7a9a441 , https://github.com/php/php-src/commit/f4e68a3968429757719c600ce16913ab09539f0d ) this fix uses new preferred way ReflectionNamedType::getName() which looks pretty stable.

### DIFF
--- a/src/DI/PhpReflection.php
+++ b/src/DI/PhpReflection.php
@@ -58,7 +58,9 @@ class PhpReflection
 	 */
 	public static function getParameterType(\ReflectionParameter $param)
 	{
-		if (PHP_VERSION_ID >= 70000) {
+		if (PHP_VERSION_ID >= 70100) {
+			return $param->hasType() ? $param->getType()->getName() : NULL;
+		} elseif (PHP_VERSION_ID >= 70000) {
 			return $param->hasType() ? (string) $param->getType() : NULL;
 		} elseif ($param->isArray() || $param->isCallable()) {
 			return $param->isArray() ? 'array' : 'callable';
--- a/tests/DI/PhpReflection.getParameterType.php71.phpt
+++ b/tests/DI/PhpReflection.getParameterType.php71.phpt
@@ -2,6 +2,7 @@
 
 /**
  * Test: Nette\DI\PhpReflection::getParameterType
+ * @phpversion >= 7.1
  */
 
 use Nette\DI\PhpReflection;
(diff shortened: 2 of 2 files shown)


src/DI/PhpReflection.php
```diff
@@ -59,7 +59,10 @@ public static function getDeclaringClass(\ReflectionProperty $prop)
 	public static function getParameterType(\ReflectionParameter $param)
 	{
 		if (PHP_VERSION_ID >= 70000) {
-			return $param->hasType() ? (string) $param->getType() : NULL;
+			if ($param->hasType()) {
+				return (PHP_VERSION_ID >= 70100) ? $param->getType()->getName() : (string) $param->getType();
+			}
+			return NULL;
```
JanTvrdik: What about using more shallow condition structure?

~~~php
if (PHP_VERSION_ID >= 70100) {
    return $param->hasType() ? $param->getType()->getName() : NULL;
} elseif (PHP_VERSION_ID >= 70000) {
    return $param->hasType() ? (string) $param->getType() : NULL;
}
~~~

tests/DI/PhpReflection.getParameterType.phpt
```diff
@@ -27,3 +27,4 @@ Assert::same('Test\B', PhpReflection::getParameterType($params[1]));
 Assert::same('array', PhpReflection::getParameterType($params[2]));
 Assert::same('callable', PhpReflection::getParameterType($params[3]));
 Assert::null(PhpReflection::getParameterType($params[4]));
+Assert::same('Test\B', PhpReflection::getParameterType($params[5]));
```
JanTvrdik: You should add `tests/DI/PhpReflection.getParameterType.php71.phpt` (with `@phpversion >= 7.1` annotation) with real nullable types.

MartinMajor:
Updated by JanTvrdik's suggestions. Thanks!

dg:
Thank you

dg:
What about return type?

MartinMajor:
Yep, it should be updated as well. I'll prepare another PR.

PR #127, opened 2016-10-12
InjectExtension: fixed injecting of services which cannot be autowired

JanTvrdik:
https://github.com/nette/di/commit/7ddd6d2f1083e3328a3b6591b8441b94dde190d6#commitcomment-19395580

Not sure if this is the right approach to solve this issue.


dg:
Seems fine, thanks.

PR #131, opened 2016-12-11
Config\Loader: allow absolute paths in includes section

milo:
I'm developing a system-wide application. Many `neon` configurations files in `/etc/app` and `/usr/lib/app/module-xxx/config.neon`. Would be more comfortable to work with absolute paths.


f3l1x:
Nice.

dg:
Thanks

milo:
Thank you!

PR #132, opened 2016-12-21
Generated factory: nullable types support

matej21:


dg:
@coveralls I have reported you as spam.

TomasVotruba:
@dg What about using Scrutinizer? It doesn't spam and doesn't report false negative on test when coverage is decreased.

dg:
@matej21 it is not a full solution, it will not work for example with `function create(?string $title, ?Foo $foo, ?int $lorem = 1): Article;`

PR #135, opened 2016-12-22
Deprecated stuff for next version [WIP]

dg:
Request for Comments @fprochazka @matej21 @mrtnzlml @JanTvrdik @hrach 

The intention is to remove stuff that complicates code while no one is using it.


mrtnzlml:
Hi @dg. I have no problem with removing inheritance support even though I think it may be useful for someone. But for some reason I have used it only in previous versions of Nette (very old) so from my point of view it's not needed.

About removing deprecated stuff: I am 100% in. I am not using `trigger_error`-ish features and in my opinion nobody does that (with new actual and stable Nette).

And new deprecated stuff: I have no problem with `parseService[s]` even though I am using it sometimes (fix is really easy). But I am sure that people are using `$this->getConfig($this->defaults)` quite often (e.g. [like this](https://github.com/Zenify/DoctrineMigrations/blob/master/src/DI/MigrationsExtension.php#L130-L131)) and I am not sure if `validateConfig` does parameter expansion? If not then it's not good replacement. I am also little bit afraid of `expand` function itself. It's very useful but `\Nette\DI\ContainerBuilder::expand` and `\Nette\DI\Container::expand` are deprecated and `\Nette\DI\Helpers::expand` is internal. Am I missing something? It's fine to use `DI\Helpers` for expansion?

dg:
ad `expand()` (and `getConfig()`): the point is to decide whether it continues to be deprecated or cancel the deprecation.

I think it's used "as smell", ie. to hide dependencies, which should not take place just in DI library.  For example, [here](https://github.com/Zenify/DoctrineMigrations/blob/master/src/DI/MigrationsExtension.php#L31) is hidden dependency for parameter `appDir`, which is not default parameter created by container. Api should not lie about depencencies, so correct way is to [declare them](https://github.com/nette/application/blob/master/src/Bridges/ApplicationDI/ApplicationExtension.php#L42).

Of course, there may exist correct usages and we can decide to unmark expand() as deprecated, or make Helpers::expand() not internal.

mrtnzlml:
I aggree with you and thanks for the point about 3rd party parameters. I didn't realize this so I am using it very often wrong (for default parameters). From my point of view is unfortunatelly very common to write something like this:

```yaml
ext:
  pathForSomething: %appDir%/path
```

And in this case it's not hiding dependencies in the extension. It's just correct usage of configuration parameters (like in services). And it's great if extension is able to understand this notation (this is what you usually expect). `ApplicationExtension` is also expanding path parameters (`scanDirs`, `scanDirs`, ...).

To conclude my ideas: I have no problem with different ways of retrieving configuration in extensions, but I would like to keep expansion possible (ie. make helper class not internal (?)). Parameter expansion is basically my only concern and I am ok with doing it manualy after config validation (just scared of relying on internal or deprecated expand feature)... :)

dg:
Now I do not know if you understand each other. Strings like `%appDir%/path` used in `config.neon` are automatically expanded before extension is triggered.

mrtnzlml:
Oh, I am very sorry. I did not realize that. If I understand correctly the expansion in the function `getConfig` is only used for expansion of it's direct parameter. If so, I am absolutelly ok with this deprecation because it doesn't make sense to expand `$this->defaults`. Thanks for your explanation.

src/DI/ServiceDefinition.php
```diff
@@ -308,33 +308,17 @@ public function getImplementMode()
...
-		trigger_error(__METHOD__ . '() is deprecated, use getImplementMode()', E_USER_DEPRECATED);
-		return $this->implementMode;
-	}
-
-
-	/** @return static */
 	public function setInject($state = TRUE)
 	{
-		//trigger_error(__METHOD__ . '() is deprecated.', E_USER_DEPRECATED);
+		trigger_error(__METHOD__ . "() is deprecated, use addTag('inject')", E_USER_DEPRECATED);
```
fprochazka: this is a nasty one :-/

dg: I have found a lot of `setInject(FALSE)` in some libraries, but have no idea why it exists…

enumag: @dg I think it's because some people used to enable inject everywhere. (very bad practice)

richard-ejem:
+1 for deprecation of all this stuff - as building container extensions is still not very intuitive, making things more clear and marking what is current best practice vs. deprecated legacy stuff is a good step.

f3l1x:
Good job @dg!

src/DI/ServiceDefinition.php
fprochazka: I actually was disabling it everywhere, to prevent it from doing some magic where I don't want it.

dg: @fprochazka :-)

Issue #130, opened 2016-12-10
Add getDefinitionByType() method


TomasVotruba:
While making extension, I often need some **service definition** in `beforeCompile()` method, e.g.:

- [load all Commands to Symfony Console Application](https://github.com/Zenify/DoctrineMigrations/blob/e82c11f8047dad44dbf16edf31afec5da636e462/src/DI/MigrationsExtension.php#L115-L122)
- [load all Fixtures to Alice](https://github.com/Zenify/DoctrineFixtures/blob/cc937219900a34ee48ee50bae08af3093ab9f06b/src/DI/FixturesExtension.php#L66-L69)
- [load all SqlFilters to Doctrine Filter Manager](https://github.com/Zenify/DoctrineFilters/blob/0c768dc78ba9e0b1c491da3170f3f1a4589d1989/src/DI/FiltersExtension.php#L42-L51)
- [load Filter Providers to Latte Engine](https://github.com/Zenify/ModularLatteFilters/blob/501b940a2a764aacebe5c781111434806e79cc3e/src/DI/ModularLatteFiltersExtension.php#L34-L41)

After making service name optional, this is the only place where it is still required.

Everywhere I need some definition, I have to use these 2 methods:

```php
$containerBuilder = $this->getContainerBuilder();
$serviceName = $containerBuilder->getByType(Nette\Bridges\ApplicationLatte\ILatteFactory::class);
$serviceDefinition = $containerBuilder->getDefinition($serviceName);
```

I would prefer having single method:

```php
$serviceDefinition = $containerBuilder->getDefinitionByType(
    Nette\Bridges\ApplicationLatte\ILatteFactory::class
);
```

What do you think?

dg:
:+1:

PR #137, opened 2017-01-04
[ContainerBuilder] getDefinitionByType() method added

TomasVotruba:
- bug fix? no #130
- new feature? yes
- BC break? no

All is described in the issue: #130 

Behavior is compatible with with `getDefiniton()` - exception is thrown when not found.
This method is called in cases where Definition is required.

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -240,6 +240,22 @@ class ContainerBuilder
 
 
 	/**
+	 * Gets the service definition of the specified type.
+	 * @param  string
+	 * @return ServiceDefinition
+	 */
+	public function getDefinitionByType($class)
+	{
+		$definitionName = $this->getByType($class);
+		if (!$definitionName) {
+			throw new MissingServiceException("Service of type '$class' not found.");
+		}
+
+		return $this->getDefinition($definitionName);
+	}
+
+
+	/**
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Ping @dg

dg:
Ping @TomasVotruba :-) https://github.com/nette/di/pull/137/files/09079155c3d878ebf5f7e42b9d4a58889978aede#diff-a46f14857c4aea27829419b9d1474390

TomasVotruba:
What need to be done here?

dg:
Can it return NULL?

TomasVotruba:
That was my first idea. But having `getDefiniton()` throwing exception and `getDefinitonByType()` giving NULL seems inconsistent.

What do you think?

dg:
So annotation `@return` should be fixed.

TomasVotruba:
I see. That is leftover from 1st idea.

Fixed!

dg:
Great, thanks!

TomasVotruba:
Awesome, thanks a lot!

TomasVotruba:
For others, here are all use cases I used it in:

https://github.com/Symplify/Symplify/pull/61/commits/1e8854c3ae5b86223fb144418f8407db831978d3

Requires `nette\di` **2.4.6+**

PR #138, opened 2017-01-05
improve auto make tmp dir func

andares:
- new feature: it looks better? if you have more path level.


JanTvrdik:
👎 This is BC break (0755 instead of default 0777), setting the 3rd parameter to TRUE is not thread-safe and is it not consistent with how other temp directories are created in Nette.

andares:
OK, I try to make it by outside logic.

PR #141, opened 2017-02-01
Compiler: loadDefinitions is called for each config separately

matej21:
- bug fix? yes
- new feature? yes
- BC break? yes

When you have multiple config files, things like alternation does not work. 
Or if you have `services: [Foo]` (without `class` key) in one config and you try to add `setup` in another config, it fails.

This may cause a bc break when a service definition in the first config is incomplete.

### DIFF
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -29,6 +29,9 @@ class Compiler
 	/** @var array */
 	private $config = [];
 
+	/** @var array */
+	private $serviceConfigs = [];
+
 	/** @var DependencyChecker */
 	private $dependencies;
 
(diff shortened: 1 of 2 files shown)


dg:
My plan is completely move config loading (and merging) from Configurator to Compiler and then learn DI to work with multiple configs. So something like this, but without BC break.

JanTvrdik:
@dg How about merging this now and refactoring it later?

dg:
@JanTvrdik It should be okay.

dg:
The `alternation` was intended as a flag that you are rewriting the service created by an compiler extension. So it should not "work" in this case. 

The problem is also with section `includes`, because configs are merged in the loader.

PR #142, opened 2017-02-02
DI\Compiler: use IExtensionsExtension while processing extensions

mrtnzlml:
This feature adds possibility to replace ExtensionsExtension with
custom implementation of IExtensionsExtension. Nette\DI\Compiler uses
this interface instead of direct implementor.

- bug fix? **no**   
- new feature? **yes**
- BC break? **no**
- doc PR: *is it needed?*

### DIFF
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -185,7 +185,7 @@ class Compiler
 		$this->config = Helpers::expand(array_diff_key($this->config, self::$reserved), $this->builder->parameters)
 			+ array_intersect_key($this->config, self::$reserved);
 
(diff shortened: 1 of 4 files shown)


dg:
IExtensionsExtension is not ideal solution, better is something like `IFirstExtension` (and `ILastExtension` for `InjectExtension`). But priorities are something that has yet to be solved, so we can add these interfaces as temporary workaround.

mrtnzlml:
Should I rename it to the `IFirstExtension` or should I wait? It doesn't have to be solved right now but I am afraid of stable version with `final` class because it cannot be workarounded easily... :)

dg:
If it doesn't have to be solved right now, you can close PR and leave issue opened.

PR #145, opened 2017-02-27
Compiler::loadDefinitions: allow to get to existing service by class

matej21:
- bug fix? no
- new feature? yes
- BC break? no

it is useful when you want to modify anonymous service in your local.neon


TomasVotruba:
Interesting. What is your specific use case?

PR #146, opened 2017-02-27
ServiceDefinition::setClass() parameter $args is deprecated (BC break)

dg:
- bug fix? no   
- new feature? no
- BC break? yes

`class` vs `factory` should be better distinguished, it is difficult to merge more configs where these options are mixed. Related to #39


JanTvrdik:
@dg Is this BC break only for compiler extensions or for *.neon files as well?

dg:
Only for direct usage of this method.

PR #147, opened 2017-03-02
Update ContainerLoader.php

andares:
add file exists check in expire check

---------------

bug fix:

use more @ tag is not a good idea in base libs. it maybe useless in some time, like use error_handler/ErrorException/exception_handler to catch error.

as a base libs, the code show be safe in most env.

### DIFF
--- a/src/DI/ContainerLoader.php
+++ b/src/DI/ContainerLoader.php
@@ -99,7 +99,11 @@ class ContainerLoader
 	private function isExpired(string $file, string &$updatedMeta = NULL): bool
 	{
 		if ($this->autoRebuild) {
-			$meta = @unserialize((string) file_get_contents("$file.meta")); // @ - file may not exist
+			$file = "$file.meta";
+			if (!file_exists($file)) {
+				return true;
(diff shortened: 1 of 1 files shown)


dg:
Your solution is not atomic and @ is perfectly fine.

JanTvrdik:
@dg I far as I can tell the solution was atomic, because he kept the original `@` in place and even though `@` works just fine, it is slightly better to avoid using it if possible, to reduce clutter of „last muted error“. 

The more annoying problem is that the proposed implementation uses unnecessary IO operation.

andares:
in fact, nobody will enable auto rebuild in production, it only work in dev or test env. atomic is not import in this scene.

dg:
The problem is that there are plenty of cases where simply you must use `@` operator. Nette uses `@` only in these cases, and usage is explained in comment. 

The clarity of „last muted error“ we can discuss as soon as PHP will offer better error handling, in the meantime, it has the lowest priority, less than increased (mini)overhead due to file_exists.

PR #151, opened 2017-03-23
In case of malformed dependencies provide type hint for calling service

MilanPala:
- bug fix? no
- new feature? no
- BC break? no

In case of malformed dependencies I got message such `"Service '94': ..."`. Now I get `"Service '94' (type of App\Presenter\HomePagePresenter): ..."`


f3l1x:
What do you think @dg about it? It would be fine merge it or reject it. I consider it useful. What about you?

PR #152, opened 2017-04-11
generated factories: fix return type

matej21:
- bug fix? yes
- new feature? no
- BC break? no

because php does not support return type covariance

### DIFF
--- a/src/DI/PhpGenerator.php
+++ b/src/DI/PhpGenerator.php
@@ -10,11 +10,10 @@ declare(strict_types=1);
 namespace Nette\DI;
 
 use Nette;
-use Nette\Utils\Validators;
 use Nette\Utils\Strings;
 use Nette\PhpGenerator\Helpers as PhpHelpers;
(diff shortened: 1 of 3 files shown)


src/DI/PhpGenerator.php
```diff
@@ -150,10 +149,12 @@ private function generateService(string $name): string
 			->addParameter('container')
 				->setTypeHint($this->className);
 
+		$rm = new \ReflectionMethod($def->getImplement(), $def->getImplementMode());
+
 		$factoryClass->addMethod($def->getImplementMode())
 			->setParameters($this->convertParameters($def->parameters))
 			->setBody(str_replace('$this', '$this->container', $code))
-			->setReturnType($def->getClass());
+			->setReturnType(Reflection::getReturnType($rm)?: $def->getClass());
```
hrach: space before ?:

dg:
Thanks!

PR #153, opened 2017-05-05
fixed including of paths with scheme in config loader

fprochazka:
- bug fix? yes
- new feature? no
- BC break? no

Including a config in the following way doesn't work

```yaml
includes:
    - phar://phpstan.phar/vendor/phpstan/phpstan-doctrine/extension.neon
```

### DIFF
--- a/src/DI/Config/Loader.php
+++ b/src/DI/Config/Loader.php
@@ -48,7 +48,7 @@ class Loader
 		if (isset($data[self::INCLUDES_KEY])) {
 			Validators::assert($data[self::INCLUDES_KEY], 'list', "section 'includes' in file '$file'");
 			foreach ($data[self::INCLUDES_KEY] as $include) {
-				if (!preg_match('#([a-z]:)?[/\\\\]#Ai', $include)) {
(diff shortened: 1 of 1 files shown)


JanTvrdik:
@fprochazka Could you add test?

fprochazka:
Probably yes, but I'm not sure how to test it. Adding a phar would be a lot of work. Maybe `file://` would be enough?

src/DI/Config/Loader.php
```diff
@@ -48,7 +48,7 @@ public function load(string $file): array
 		if (isset($data[self::INCLUDES_KEY])) {
 			Validators::assert($data[self::INCLUDES_KEY], 'list', "section 'includes' in file '$file'");
 			foreach ($data[self::INCLUDES_KEY] as $include) {
-				if (!preg_match('#([a-z]:)?[/\\\\]#Ai', $include)) {
+				if (!preg_match('#([a-z]+:)?[/\\\\]#Ai', $include)) {
```
Majkl578: This will also allow i.e. `phar:\\` which is not correct.

Issue #156, opened 2017-07-06
Bug with class_alias


enumag:
- bug report? yes
- feature request? no
- version: 2.4.8

### Description
Nette/DI seems to be incompatible with class aliases. My tests are failing with this error:

```
[Nette\DI\ServiceCreationException] Service 'arachne.forms.twig.extension.translation': Case mismatch on class name 'Twig\NodeVisitor\NodeVisitorInterface', correct name is 'Twig_NodeVisitorInterface'.
```

Obviously there is no case mismatch. It's just that `Twig\NodeVisitor\NodeVisitorInterface` is an [alias](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/NodeVisitorInterface.php#L43) of `Twig_NodeVisitorInterface` which confuses [ContainerBuilder::checkCase()](https://github.com/nette/di/blob/03d0a601f3bb4ceb045ded4771ff07c21d1dcc79/src/DI/ContainerBuilder.php#L700).

### Steps To Reproduce
Create a class and an alias to that class. Add it to DIC as service. Try to autowire it into another service using the alias instead of the base class. EDIT: See the test case in my PR.

enumag:
Not sure if it's possible to fix it otherwise than ignoring it with

```php
	private function checkCase($class)
	{
		if (class_exists($class) || interface_exists($class)) {
			$name = (new ReflectionClass($class))->getName();
			if ($class !== $name && strtolower($class) === strtolower($name)) {
				throw new ServiceCreationException("Case mismatch on class name '$class', correct name is '$name'.");
			}
		}
	}

```

enumag:
@dg Any estimate 2.4.9 with this fix could be available? This bug is blocking some of my open-source projects.

dg:
@enumag done

enumag:
Thank you!

PR #157, opened 2017-07-06
ContainerBuilder: Fix bug with class alias

enumag:
- bug fix? yes #156
- new feature? no
- BC break? no
- docs PR? not needed


enumag:
The coveralls failure it false-positive. This PR does not decrease code coverage.

PR #158, opened 2017-08-14
DI config: add spread operator

matej21:
- bug fix? no
- new feature? yes
- BC break? yes (when someone concats `...` with a scalar parameter. it could be fixed by checking whether return value is an array, but is it worth it?)


dg:
For accuracy: it is not like PHP spread operator, because it works with the keys, so `DbConnection(...%connection%)` in NEON and `new DbConnection(...$connection)` in PHP may work different.

f3l1x:
It would be fine merge it or reject it. What do you think? I could be the fine feature in Nette 3?

PR #160, opened 2017-09-29
Add deprecation notes

enumag:


dg:
Thanks

enumag:
@dg Can you cherry pick this to 2.4 pls? I got another newbie today asking what to use instead of getClass.

dg:
Done

PR #165, opened 2018-02-14
 added getParameter method

asmarcz:
- bug fix? no
- new feature? yes
- BC break? no

It sounds logical to me to use for example `$this->context->getParameter('wwwDir')` instead of `$this->context->getParameters()['wwwDir']`.


dg:
You can use `$this->context->parameters['wwwDir']`, it is public property.

PR #167, opened 2018-07-04
Container::getByType fixed phpdoc return typehint

CZechBoY:
- bug fix
- BC break? no

Container::getByType really returns null (previously "returned" void).

### DIFF
--- a/src/DI/Container.php
+++ b/src/DI/Container.php
@@ -183,7 +183,7 @@ class Container
 	/**
 	 * Resolves service by type.
 	 * @param  bool  $throw  exception if service doesn't exist?
-	 * @return object  service or null
+	 * @return object|null service
 	 * @throws MissingServiceException
 	 */
 	public function getByType(string $type, bool $throw = true)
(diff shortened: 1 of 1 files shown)


src/DI/Container.php
```diff
@@ -251,7 +253,7 @@ public function createInstance(string $class, array $args = [])
 
 	/**
 	 * Calls all methods starting with with "inject" using autowiring.
-	 * @param  object  $service
+	 * @param object $service
```
dg: This was ok, or not?

CZechBoY: yes, I can revert it if you want... or split to 2nd commit

CZechBoY: splitted to 2nd commit and removed some other unnecessary spaces in phpdocs

dg: These spaces are used intentionally and across all libraries.

CZechBoY: Ok, I don't know why to add unnecessary spaces, but it is your opinion. Commit reverted.

enumag:
Oh great, now PHPStan will fail on every single usage of `Container::getByType()` because it will think that the method can return null. This is why boolean $throw parameters are evil.

CZechBoY:
@enumag I think there is return type extension already... Do you have phpstan/phpstan-nette included?

PR #168, opened 2018-07-26
DIExtension: service name tagged with run cast to string

MartkCz:
- bug fix  
- BC break? no

Following code 
```yaml
services:
    - {factory: stdClass, tags: [run]}
```
causes TypeError: ` TypeError: Argument 1 passed to Nette\DI\Container::getService() must be of the type string, integer given`


dg:
Thanks

PR #173, opened 2018-08-08
DependencyChecker: added callback support

adawolfa:
This PR adds possibility to invalidate the DI container using given callback.

- **feature**
- **not aware of any BC break**

Currently, you can make `DependencyChecker` track specific files (and classes or methods). Using callback, we could invalidate the container, for instance:

- when there is a new file (like a presenter with routing annotation - this is actually real use case),
- when *something* of environment changes.

I'm not sure about proper implementation - open to discussion. What concerns me is `isExpired()` method; it used to call `calculateHash()` only when a file was changed, but now it gets called always when there is a callback as well (which causes tracked files to be loaded and inspected unnecessarily). An optimization would require a separate hash for callbacks.

## Usage

Dependency is supposed to be defined like this:

~~~
Array(string & callable $callback, mixed ...$arg)
~~~

First argument must be a `callable` and `string`, which limits the callback to a function or a static method. Remaining arguments are stored within the callback and are always passed to it (I can explain what's that good for if it's not obvious).

~~~php
$compiler->addDependencies([

    // track for PHP version change
    ['phpversion'],

    // track for an extension
    ['extension_loaded', 'sockets'],

    // stupid simple file watcher
    ['glob', 'src/*/*Presenter.php'],

    // custom callback
    ['MyClass::getInstallationHash'],

]);
~~~


milo:
I would consider sytanx for static method call as `[MyClass::class, 'method']`. It is easier to maintain in IDE. Sure, uglier version `[MyClass::class . '::method']` would work ;)

adawolfa:
Yup, here's what I have been thinking:

- the array syntax is mostly used with an object as a first argument - we'd have to check and reject such callback,
- if we use the array syntax, how arguments will be passed (can't think of an intuitive way)?

I don't really mind anyway, if you want, I'll edit the PR.

milo:
Probably `[[MyClass::class, 'method'], 'arg', 'arg']` but save your time and don't edit it now. It's just note.

JanTvrdik:
(Also note that support for `MyClass::class . '::method'` MAY be [removed](https://wiki.php.net/rfc/consistent_callables#remove_colon_separated_string_callables_-_8) in PHP 8)

adawolfa:
I've allowed the callback to be specified using array syntax.

Something about performace - would it be OK to use part of hash for files and other part for callbacks? I think there's no need for entirely separated hash.

dg:
You think Callback::isStatic() may help https://api.nette.org/2.4/Nette.Utils.Callback.html

dg:
> Something about performace - would it be OK to use part of hash for files and other part for callbacks? I think there's no need for entirely separated hash.

It must be separated. Hashes for callbacks will be generated for each request and it should not influence generating hashes for files.

adawolfa:
Updates:
- files and callbacks are checked individually using single concatenated hash,
- no failure on package update with cached container,
- using `Nette\Utils\Callback::isStatic()`.

src/DI/DependencyChecker.php
```diff
@@ -70,28 +73,29 @@ public function export(): array
 
 		$classes = array_keys($classes);
 		$functions = array_unique($functions, SORT_REGULAR);
-		$hash = self::calculateHash($classes, $functions);
+		$hash = self::calculateReflectionHash($classes, $functions) . self::calculateCallbackHash($callbacks);
 		$files = @array_map('filemtime', array_combine($files, $files)); // @ - file may not exist
 		$phpFiles = @array_map('filemtime', array_combine($phpFiles, $phpFiles)); // @ - file may not exist
-		return [self::VERSION, $files, $phpFiles, $classes, $functions, $hash];
+		return [self::VERSION, $files, $phpFiles, $classes, $functions, $callbacks, $hash];
```
dg: Maybe simpler solution will be better, ie. `$reflectionHash = self::calculateReflectionHash(); $callbackHash = self::calculateCallbackHash();` and then `return [self::VERSION, $files, $phpFiles, $classes, $functions, $callbacks, $reflectionHash, $callbackHash]`, what do you mean?

adawolfa: Yes, that's was original solution (linked above). I'll split them, if you wish. How about compatibility with cached container, should I drop it or keep it?

dg: Compatibility is not needed, ie. it is enough to have compatible isExpired declaration:

```php
public static function isExpired(int $version, array $files, array &$phpFiles, array $classes, array $functions, string $reflectionHash, array $callbacks = [], string $callbackHash = '')
```

dg:
That makes sense, but it must be done a little differently. I'll try to see it later.

PR #176, opened 2018-09-15
ContainerBuilder: set caused by exception

CZechBoY:
- bug fix? yes
- BC break? no

Added link to previous exception when service creation failed due to reflection exception - because for example class not found.

### DIFF
--- a/src/DI/ContainerBuilder.php
+++ b/src/DI/ContainerBuilder.php
@@ -487,7 +487,7 @@ class ContainerBuilder
 			if (isset($e) || ($refClass && (!$reflection->isPublic()
 				|| ($refClass->isTrait() && !$reflection->isStatic())
 			))) {
-				throw new ServiceCreationException(sprintf("Method %s() used in service '%s' is not callable.", Nette\Utils\Callback::toString($entity), $serviceName));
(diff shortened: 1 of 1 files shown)


src/DI/ContainerBuilder.php
```diff
@@ -487,7 +487,7 @@ private function resolveEntityType($entity, array $recursive = []): ?string
 			if (isset($e) || ($refClass && (!$reflection->isPublic()
 				|| ($refClass->isTrait() && !$reflection->isStatic())
 			))) {
-				throw new ServiceCreationException(sprintf("Method %s() used in service '%s' is not callable.", Nette\Utils\Callback::toString($entity), $serviceName));
+				throw new ServiceCreationException(sprintf("Method %s() used in service '%s' is not callable.", Nette\Utils\Callback::toString($entity), $serviceName), isset($e) ? $e->getCode() : 0, $e ?? null);
```
dg: I think that code should not be copied, because it is code for different exception.

TomasVotruba:
Nice!

I found similar missed exception with this PHPStan set: http://bestpractices.thecodingmachine.com/php/error_handling.html#wrapping-an-exception-do-not-lose-the-previous-exception, https://github.com/thecodingmachine/phpstan-strict-rules/blob/master/README.md

Useful for full code base changes

PR #178, opened 2018-10-03
autowiring via Service[] (WIP)

dg:
Autowiring of array of `Service` this way: (BC break)

```php
class Foo
{
	public $services;

	/**
	 * @param Service[] $services
	 */
	public function __construct(array $services)
	{
		$this->services = $services;
	}
}
```

### DIFF
--- a/src/DI/Autowiring.php
+++ b/src/DI/Autowiring.php
@@ -189,6 +189,16 @@ class Autowiring
 					$optCount = 0;
 				}
 
+			} elseif (
+				$container instanceof Resolver
+				&& $method instanceof \ReflectionMethod
+				&& $parameter->isArray()
+				&& preg_match('#@param[ \t]+([\w\\\\]+)\[\][ \t]+\$' . $paramName . '#', (string) $method->getDocComment(), $m)
+				&& ($type = Reflection::expandClassName($m[1], $method->getDeclaringClass()))
+				&& (class_exists($type) || interface_exists($type))
+			) {
+				$res[$num] = $container->findByType($type);
+
 			} elseif (($type && $parameter->allowsNull()) || $parameter->isOptional() || $parameter->isDefaultValueAvailable()) {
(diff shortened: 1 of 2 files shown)


milo:
This is definetly time saver!

milo:
Without this (and previous `tags() & types()` commit), I'm solving it by extension. I have collections like:
```php
class MyCollection
{
    public function add(MyInterface $service) {...}
}
```
and [this DI extension](https://gist.github.com/milo/94313f1cf9a5c4c499cd1cd173df5984) which, in a short, finds all services of `MyInterface` type in container and creates `MyCollection` service with `add()` calls.

With proposed syntax, I don't need the extension anymore, because this collections have static factory `createFromArray(array $items)` too 👍

enumag:
I'm solving it with this [package](https://github.com/Arachne/ServiceCollections) which is much more powerful. This only solves the most simple use-case so it's not nearly enough for me.

dg:
@enumag Can you describe what it does?

dg:
Point 2 looks useful, it is good idea to have multiple form of accessor.

```php
interface DbAccessor
{
	function get(): Db; 
	// and allow this too:
	function get($name): Db
}
```

There should be enumeration of services in configuration, ie.:

```neon
services:
	- DbAccessor(
		name1: @a
		name2: @b
	)
	a: Db('***')
	b: Db('***')
	c: Db('***')
```

enumag:
Personally I prefer configuration with tags (the name is a tag attribute) instead of inventing some new syntax but that's a detail.

Which service would the `function get(): Db;` return? I don't see a reason to allow calling it without argument...

dg:
Currently there si supported this syntax

```neon
services:
	- DbAccessor(@a)
```

so it is only enhancement. Tags seems like another way.

enumag:
Oh, right, forgot about that. Yeah this seems like a good improvement. 👍

PR #185, opened 2019-01-25
Processor: replaced call of depracated getFactory() method on definition

Daaarkling:
- bug fix
- BC break? no

Hi, I have found bug in beta version. When you try to use config like shown below, it throws "User Deprecated" error.

```neon
services:
	- 
		implement: SomeInterface
		arguments:
			foo: bar
```


dg:
thanks

PR #189, opened 2019-02-22
Fixed return annotation in getResultDefinition method

michalhlavka:
- bug fix?   yes
- BC break? no


dg:
This is intention due to the better code completion in PhpStorm.

JanTvrdik:
how about `@return Definition|ServiceDefinition`?

PR #191, opened 2019-03-12
Schema for unified automatic configuration validation 

dg:
- BC break? no

CompilerExtension can provide a schema which is automatically used to validate and normalize configuration:

```php
	public function getConfigSchema(): Nette\DI\Config\Schema
	{
		return Expect::arrayOf([
			'setup' => Expect::list(),
			'tags' => Expect::array(),
			'inject' => Expect::bool(),
		]);
	}
```


holantomas:
Great! Only one thing. Did you think about something like `Expect::string()->nullable();` instead of using `enum()`? If you look on database schema it's look little bit confusing(for me).

```php
Expect::arrayOf([
			'dsn' => Expect::string()->required(), // Have to be presented, but CAN be null?
			'user' => Expect::enum(Expect::string(), null), // Have not to be presented and CAN be null
			'reflection' => Expect::string(), // Have not to be presented but CANNOT be null?
			//...
		])
```

Maybe this look better
```php
Expect::arrayOf([
			'dsn' => Expect::string()->required(), // Have to be present and CANNOT be null
			'user' => Expect::string()->nullable(), // Have not be presented and CAN be null
			'reflection' => Expect::string(), // Have not to be presented but CANNOT be null
			'password' => Expect::string()->required()->nullable(), // Have to be presented but CAN be null
			//...
		])
```

Or am I missing something?

dakorpar:
@dg why not making popo for config, that way we're far more strict and full autocomplete in all ides, and with php7.4 we will have typehints on properties? Wouldn't that be far better then this dynamic validation?

src/DI/Config/Expect.php
```diff
@@ -0,0 +1,312 @@
...
+
+	public static function __callStatic(string $name, array $args): self
+	{
+		$me = new self($name);
+		$me->default = $args[0] ?? $me->default;
+		return $me;
+	}
+
+
+	public static function type(string $type): self
```
milo: What about:
```php
function type(string ...$types): self
{
    return new self(implode('|', $types));
}
```
To be able write:
```php
Expect::type('string', Nette\DI\Definitions\Statement::class)
```
or own method for it.

dg:
@dakorpar I also thought about using the class for configuration and it's definitely a good idea.

dg:
image

dg:
image

ondrejmirtes:
Cool! Can I somehow use the system to validate what users have in their `parameters` section?

dg:
@ondrejmirtes of course!

dg:
@holantomas I added `nullable()` and `require()`, thanks for suggestions

trejjam:
Hi, thanks for this I really like it!
Is there any recomemded way how to define schema for a pre-v3 compatible nette extensions?
My point is that when I define `getConfigSchema(): Nette\DI\Config\Schema` method in extensions class, php compiler will yell at me that it does not know class `Nette\DI\Config\Schema`.

JanTvrdik:
> php compiler will yell at me

Are you sure? Afaik return types to not trigger autoloading.

trejjam:
I am not sure, thanks. It was a habit from compiled languages.

PR #197, opened 2019-04-29
Added Decorator extension to $last

peldax:
- Bug fix
- BC break? no

Decorator extension needs to be resolved last (eg. after Search extension) in order to find all possible matching services.

Solves my issue where decorator doesn't decorate services found using Search extension.


vossik:
Ping @dg

dg:
Thanks

Issue #196, opened 2019-04-26
Search Extension - search for factory interfaces


peldax:
Hi,
I really love search extension in 3.0. I removed all the long config files with model classes I used as services and replaced it with few lines of search pattern.

I would love to be able to search for Component factories as well, eg. classes like:
```
<?php

declare(strict_types = 1);

namespace App\XYZ\Component;

interface IXYZFactory
{
    public function create() : XYZ;
}
```
They are currently listed in config file as following:
```
services:
    - App\XYZ\Component\IXYZFactory
```

Maybe its already possible, but I couldn't get it to work. From my inspection of code, there is a condition for searched class to be instantiable, which interface isn't, so I believe the extension need to be modified.

From quick inspection of how services extension work I couldn't quite find any solution, so I am leaving this request here. If you could give me some pointers what needs to be done I may send a PR.

Thanks.
VP

PR #200, opened 2019-05-10
SearchExtension - ability to create factory definitions from interfaces

peldax:
- new feature  #196 
- BC break? no

This PR extends SearchExtension to be able to search for interfaces and create factory definitions from them.

WIP is added due to some duplicate code that needs to be refactored.


peldax:
@dg I am making tests for this PR and I am wordering what is desirable outcome when factory interface doesnt implement the `create()` function
- Skip the interface silently?
- Throw exception that interface doesnt have the function?

PR #204, opened 2019-07-13
Generate cached container file independently of class existence

mabar:
- bug fix
- BC break? no
- doc PR: not needed

I have a build reload command which deletes cache and then generate all possible container configurations (console, debug, production, ...). During cache removal is deleted also currently running container so class exists while it's file not and I need to create it again

### DIFF
--- a/src/DI/ContainerLoader.php
+++ b/src/DI/ContainerLoader.php
@@ -40,9 +40,7 @@ class ContainerLoader
 	public function load(callable $generator, $key = null): string
 	{
 		$class = $this->getClassName($key);
-		if (!class_exists($class, false)) {
-			$this->loadFile($class, $generator);
-		}
+		$this->loadFile($class, $generator);
 		return $class;
 	}
 


dg:
I do not understand. You can't load one class twice.

mabar:
I am not trying load one class twice. I am trying to generate it twice. During complete app reinstall is cached container deleted. Cache is warmed up then but file is not regenerated because container class still exists.

It should not break anything, loadFile() checks for file existence https://github.com/nette/di/blob/master/src/DI/ContainerLoader.php#L62-L64

dg:
> I am not trying load one class twice

It will call `@include $file` twice, or not?

mabar:
You are right. I missed the second include after file generation. It should be ok now.

dg:
I understand what you want to do, but I don't like this way. Now don't make sense method names like `loadFile()`, behavior of `load()`, etc. This requires more sophisticated refactoring.

mabar:
I would appreciate help then. I don't know how to write it better

dg:
I don't know either, of course. It takes more to think about. Maybe split into two classes… Because what you want is not task for the "loader".

mabar:
Generator sounds acceptable. I will take a look at it later

mabar:
Looking at it after some time, bad idea... solved by ommiting currect container when deleting generated files.

PR #206, opened 2019-07-18
Remove redundant $service variable in case of simply service without dependencies.

janbarasek:
In case of:

```php
public function createServiceMyService(): Model\MyService
{
	$service = new Model\MyService;
	return $service;
}
```

Generate simplify version:

```php
public function createServiceMyService(): Model\MyService
{
	return new Model\MyService;
}
```

And in case of simple static service:

```php
public function createServiceMyService(): Model\MyService
{
	$service = Model\MyServiceFactory::getInstance();
	return $service;
}
```

Generate:

```
public function createServiceMyService(): Model\MyService
{
	return Model\MyServiceFactory::getInstance();
}
```

### DIFF
--- a/src/DI/PhpGenerator.php
+++ b/src/DI/PhpGenerator.php
@@ -90,6 +90,9 @@ declare(strict_types=1);
 			$method->setVisibility('public');
 			$method->setReturnType($def->getType());
 			$def->generateMethod($method, $this);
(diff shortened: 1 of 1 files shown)


dg:
I think it would be cleaner to deal with this in `ServiceDefinition::generateMethod()`…

PR #209, opened 2019-07-18
ExtensionsExtension: Fix type hint of RC

janbarasek:
- bug fix
- BC break? no

Fix type hint and solve error:

```
------ --------------------------------------------------------------------- 
  Line   DI/Extensions/ExtensionsExtension.php                                
 ------ --------------------------------------------------------------------- 
  34     Parameter #2 $extension of method Nette\DI\Compiler::addExtension()  
         expects Nette\DI\CompilerExtension, object given.                    
 ------ --------------------------------------------------------------------- 
```

### DIFF
--- a/src/DI/Extensions/ExtensionsExtension.php
+++ b/src/DI/Extensions/ExtensionsExtension.php
@@ -30,8 +30,9 @@ final class ExtensionsExtension extends Nette\DI\CompilerExtension
 				$name = null;
 			}
 			if ($class instanceof Nette\DI\Definitions\Statement) {
-				$rc = new \ReflectionClass($class->getEntity());
-				$this->compiler->addExtension($name, $rc->newInstanceArgs($class->arguments));
+				/** @var Nette\DI\CompilerExtension $rcExtension */
+				$rcExtension = (new \ReflectionClass($class->getEntity()))->newInstanceArgs($class->arguments);
+				$this->compiler->addExtension($name, $rcExtension);
 			} else {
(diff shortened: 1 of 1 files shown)


src/DI/Extensions/ExtensionsExtension.php
```diff
@@ -30,8 +30,9 @@ public function loadConfiguration()
 				$name = null;
 			}
 			if ($class instanceof Nette\DI\Definitions\Statement) {
-				$rc = new \ReflectionClass($class->getEntity());
-				$this->compiler->addExtension($name, $rc->newInstanceArgs($class->arguments));
+				/** @var Nette\DI\CompilerExtension $rcExtension */
```
dg: In fact, we don't know if the `$rcExtension` is CompilerExtension, and it is expected that it might throw a TypeError.

mabar: Btw, it is better to use `assert($rcExtension instanceof CompilerExtension)`
Locally defined annotations are hack to satisfy IDE

dg: Ideally, there should be checking and throwing an exception.

mabar: `assert` should be used instead of local annotations for internal behavior, to be exact.

janbarasek: I implemented exception, but I'm not sure about the message.

PR #210, opened 2019-07-18
InjectExtension: Fix type hint and always executed if.

janbarasek:
- bug fix
- BC break? no

Fix type error and if condition order:

```
 ------ ---------------------------------------------------------- 
  Line   DI/Extensions/InjectExtension.php                         
 ------ ---------------------------------------------------------- 
  37     Right side of && is always true.                          
  61     Parameter #4 $container of static method                  
         Nette\DI\Extensions\InjectExtension::checkType() expects  
         Nette\DI\Container|Nette\DI\Resolver|null,                
         Nette\DI\ContainerBuilder|null given.                     
 ------ ---------------------------------------------------------- 
```


dg:
Thanks

PR #212, opened 2019-07-18
PhpGenerator: $name might not be defined

janbarasek:
- bug fix
- BC break? no

Solve:

```
 ------ -------------------------------------- 
  Line   DI/PhpGenerator.php                   
 ------ -------------------------------------- 
  99     Variable $name might not be defined.  
 ------ -------------------------------------- 
```


f3l1x:
Hi. It is great you improving nette/di. I would suggest next time to join all changes to one PR, further codereview will be much easier.

janbarasek:
@f3l1x I am very sorry for it. I thought it was better to check all the micro-fixations separate.

dg:
Thanks

PR #213, opened 2019-07-18
Resolver: Fix undefined and never called spaces.

janbarasek:
- bug fix
- BC break? no

Fix and basic reimplementation for define type hints in comment section and little code reorganization.

```
 ------ --------------------------------------------------------------------- 
  Line   DI/Resolver.php                                                      
 ------ --------------------------------------------------------------------- 
  117    Call to an undefined method ReflectionFunctionAbstract::isPublic().  
  117    Variable $refClass might not be defined.                             
  117    Variable $reflection might not be defined.                           
  118    Call to an undefined method ReflectionFunctionAbstract::isStatic().  
  118    Variable $reflection might not be defined.                           
  122    Variable $reflection might not be defined.                           
  124    Variable $reflection might not be defined.                           
  335    Negated boolean expression is always false.                          
 ------ --------------------------------------------------------------------- 
```

### DIFF
--- a/src/DI/Resolver.php
+++ b/src/DI/Resolver.php
@@ -108,12 +108,12 @@ class Resolver
 				}
 			}
 
-			/** @var \ReflectionMethod $reflection */
-			$reflection = Nette\Utils\Callback::toReflection($entity[0] === '' ? $entity[1] : $entity);
-
 			try {
+				/** @var \ReflectionMethod $reflection */
+				$reflection = Nette\Utils\Callback::toReflection($entity[0] === '' ? $entity[1] : $entity);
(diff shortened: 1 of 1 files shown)


src/DI/Resolver.php
```diff
@@ -319,23 +326,19 @@ public function normalizeReference(Reference $ref): Reference
...
-			}
-			if (!$res) {
-				throw new ServiceCreationException("Reference to missing service of type $service.");
-			}
-			return $res;
+		}
+		try {
+			return $this->getByType($service);
+		} catch (NotAllowedDuringResolvingException $e) {
+			return new Reference($service);
```
dg: I feel like this changes the meaning of the code, there is missing exception.

PR #214, opened 2019-07-24
PhpStan fixes

janbarasek:
- bug fix
- BC break? yes

Some more repairs by PhpStan.

### DIFF
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -149,7 +149,15 @@ class Compiler
 	 */
 	public function setDynamicParameterNames(array $names)
 	{
-		$this->extensions[self::PARAMETERS]->dynamicParams = $names;
+		$extension = $this->extensions[self::PARAMETERS];
+
+		if ($extension instanceof Nette\DI\Extensions\ParametersExtension) {
+			$extension->dynamicParams = $names;
+		} else {
+			throw new InvalidConfigurationException(
+				"Parameter 'dynamicParams' in extension '" . self::PARAMETERS . "' does not exist."
+			);
+		}
 		return $this;
 	}
 
@@ -180,8 +188,14 @@ class Compiler
 	 */
 	public function addExportedTag(string $tag)
 	{
-		if (isset($this->extensions[self::DI])) {
-			$this->extensions[self::DI]->exportedTags[$tag] = true;
+		$extension = $this->extensions[self::DI];
+
+		if ($extension instanceof Nette\DI\Extensions\DIExtension) {
+			$extension->exportedTags[$tag] = true;
+		} else {
+			throw new InvalidConfigurationException(
+				"Parameter 'exportedTags' in extension '" . self::DI . "' does not exist."
+			);
 		}
 		return $this;
 	}
@@ -192,8 +206,14 @@ class Compiler
 	 */
 	public function addExportedType(string $type)
 	{
-		if (isset($this->extensions[self::DI])) {
-			$this->extensions[self::DI]->exportedTypes[$type] = true;
+		$extension = $this->extensions[self::DI];
+
+		if ($extension instanceof Nette\DI\Extensions\DIExtension) {
+			$extension->exportedTypes[$type] = true;
+		} else {
+			throw new InvalidConfigurationException(
+				"Parameter 'exportedTypes' in extension '" . self::DI . "' does not exist."
+			);
 		}
 		return $this;
 	}
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -150,14 +150,8 @@ class Compiler
 	public function setDynamicParameterNames(array $names)
 	{
 		$extension = $this->extensions[self::PARAMETERS];
-
-		if ($extension instanceof Nette\DI\Extensions\ParametersExtension) {
-			$extension->dynamicParams = $names;
-		} else {
-			throw new InvalidConfigurationException(
-				"Parameter 'dynamicParams' in extension '" . self::PARAMETERS . "' does not exist."
-			);
-		}
+		assert($extension instanceof Nette\DI\Extensions\ParametersExtension);
+		$extension->dynamicParams = $names;
 		return $this;
 	}
 
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -183,14 +183,8 @@ class Compiler
 	public function addExportedTag(string $tag)
 	{
 		$extension = $this->extensions[self::DI];
-
-		if ($extension instanceof Nette\DI\Extensions\DIExtension) {
-			$extension->exportedTags[$tag] = true;
-		} else {
-			throw new InvalidConfigurationException(
-				"Parameter 'exportedTags' in extension '" . self::DI . "' does not exist."
-			);
-		}
+		assert($extension instanceof Nette\DI\Extensions\DIExtension);
+		$extension->exportedTags[$tag] = true;
 		return $this;
 	}
 
@@ -201,14 +195,8 @@ class Compiler
 	public function addExportedType(string $type)
 	{
 		$extension = $this->extensions[self::DI];
-
-		if ($extension instanceof Nette\DI\Extensions\DIExtension) {
-			$extension->exportedTypes[$type] = true;
-		} else {
-			throw new InvalidConfigurationException(
-				"Parameter 'exportedTypes' in extension '" . self::DI . "' does not exist."
-			);
-		}
+		assert($extension instanceof Nette\DI\Extensions\DIExtension);
+		$extension->exportedTypes[$type] = true;
 		return $this;
 	}
 
--- a/src/DI/Compiler.php
+++ b/src/DI/Compiler.php
@@ -182,9 +182,10 @@ class Compiler
 	 */
 	public function addExportedTag(string $tag)
 	{
-		$extension = $this->extensions[self::DI];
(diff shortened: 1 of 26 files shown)


src/DI/Compiler.php
```diff
@@ -149,7 +149,15 @@ public function getConfig(): array
...
-		$this->extensions[self::PARAMETERS]->dynamicParams = $names;
+		$extension = $this->extensions[self::PARAMETERS];
+
+		if ($extension instanceof Nette\DI\Extensions\ParametersExtension) {
+			$extension->dynamicParams = $names;
+		} else {
+			throw new InvalidConfigurationException(
+				"Parameter 'dynamicParams' in extension '" . self::PARAMETERS . "' does not exist."
+			);
+		}
```
mabar: `assert($extension instanceof Nette\DI\Extensions\ParametersExtension)` is better for internal behavior. This exception can never happen and assert is enough to satisfy IDE and also phpstan.

src/DI/Definitions/FactoryDefinition.php
```diff
@@ -231,6 +231,13 @@ private function completeParameters(Nette\DI\Resolver $resolver): void
 		$interface = $this->getType();
 		$method = new \ReflectionMethod($interface, self::METHOD_CREATE);
 
+		if (!$this->resultDefinition instanceof ServiceDefinition) {
+			throw new ServiceCreationException(
+				'Result definition must be type of ServiceDefinition, '
+				. "type '" . get_class($this->resultDefinition) . "' given."
```
mabar: This cannot work. `setResultDefinition()` allows `Definition`. I would personally allow only `ServiceDefinition`, but I guess it's an intention and most common return type is defined in annotation just because generics support is not good enough yet.
cc @dg

src/DI/ContainerBuilder.php
```diff
@@ -55,16 +55,17 @@ public function __construct()
...
 	public function addDefinition(?string $name, Definition $definition = null): Definition
 	{
 		$this->needsResolve = true;
+		$nameType = $name === null ? null : key([$name => 1]); /** @var string|int|null $nameType */
 		if ($name === null) {
 			for ($i = 1; isset($this->definitions['0' . $i]) || isset($this->aliases['0' . $i]); $i++);
 			$name = '0' . $i; // prevents converting to integer in array key
 
-		} elseif (is_int(key([$name => 1])) || !preg_match('#^\w+(\.\w+)*$#D', $name)) {
+		} elseif (is_int($nameType) || !preg_match('#^\w+(\.\w+)*$#D', $name)) {
```
mabar: @dg This `key()` usage seems like hacky way to check if string is numeric and I think auto-conversion will stop working with some strict directive in future php versions. Maybe `Nette\Utils\Validators::isNumeric|isNumericInt()` would work here.

src/DI/Compiler.php
dg: does phpstan also understand this? `assert($this->extensions[self::PARAMETERS] instanceof Nette\DI\Extensions\ParametersExtension)`

src/DI/ContainerBuilder.php
dg: Yes, it is intentional.

src/DI/Compiler.php
mabar: I am not sure. Edge case which I never needed.

src/DI/Definitions/FactoryDefinition.php
```diff
@@ -74,7 +74,7 @@ public function setResultDefinition(Definition $definition)
 
 
 	/**
-	 * @return ServiceDefinition
+	 * @return Definition|ServiceDefinition
```
dg: ServiceDefinition is Definition

src/DI/Compiler.php
dg: I tried it and it works.

src/DI/Definitions/FactoryDefinition.php
mabar: `@return ServiceDefinition` -> returns ServiceDefinition only

`@return Definition|ServiceDefinition` -> returns Definition or it's child ServiceDefinition. Definition only would make most sense here, but if we consider ServiceDefinition is almost everytime used then it's nice to satisfy IDE.

But I guess phpstan will start reporting invalid calls to methods which are on ServiceDefinition only. Imho better not to fix, just add it to ignore.

src/DI/ContainerBuilder.php
```diff
@@ -55,16 +55,17 @@ public function __construct()
 
 	/**
 	 * Adds new service definition.
-	 * @return Definitions\ServiceDefinition
+	 * @return Definition|Definitions\ServiceDefinition
```
dg: Definition is subclass of ServiceDefinition

src/DI/ContainerBuilder.php
dg: Maybe it should be fixed in PhpStan, key() can return int.

mabar: It can, but I think problem here is in `key()` used to convert numeric value to int, it's a side-effect.

src/DI/Definitions/FactoryDefinition.php
dg: What I mean is that it would be enough to write `Definition`.

And yes, it really exists only to satisfy IDE. But it is not correct, of course.

src/DI/ContainerBuilder.php
mabar: I tried to change it and got reported `Call to an undefined method Nette\DI\Definitions\Definition::setFactory().`. Which is true, but this would break every single repo using phpstan. Better not to fix, same as for change in FactoryDefinition

src/DI/ContainerBuilder.php
dg: It's also the surest way to verify that the key will not be number.

src/DI/ContainerBuilder.php
mabar: Dark side of a BC :/ 
`addDefinition(Definition $definition, ?string $name = null): void` would work perfectly.

dg: I agree that this should not be removed until a solution for 3rd code is available.

src/DI/ContainerBuilder.php
JanTvrdik: `key()` is definitely bug in phpstan, caused by combination of https://github.com/phpstan/phpstan/blob/3ef2c05f0fc5dbe413674ef33cbbd60c33bb1ce7/src/Type/Php/ArrayKeyDynamicReturnTypeExtension.php and `ArrayType::getIterableKeyType()` being incorrect because it prefer usefulness over accuracy.

src/DI/Definitions/FactoryDefinition.php
janbarasek: Yes, true. I reverted old state.

src/DI/Definitions/FactoryDefinition.php
janbarasek: Yes, I understand. I reverted old state.

src/DI/ContainerBuilder.php
janbarasek: I suggest fix it like in `Nette\Utils\Validators::isNumericInt()`, specifically: `($name !== null && preg_match('#^[+-]?\d+\z#', $name))`

src/DI/ContainerBuilder.php
janbarasek: Removed.

src/DI/ContainerBuilder.php
janbarasek: Ok, I am very sorry. Correct pattern is `preg_match('#^[+-]?([1-9]\d*|0)\z#', $name)`.

src/DI/Compiler.php
janbarasek: Fixed.

src/DI/Compiler.php
```diff
@@ -180,8 +182,14 @@ public function exportDependencies(): array
...
-		if (isset($this->extensions[self::DI])) {
-			$this->extensions[self::DI]->exportedTags[$tag] = true;
+		$extension = $this->extensions[self::DI];
+
+		if ($extension instanceof Nette\DI\Extensions\DIExtension) {
+			$extension->exportedTags[$tag] = true;
+		} else {
+			throw new InvalidConfigurationException(
+				"Parameter 'exportedTags' in extension '" . self::DI . "' does not exist."
+			);
```
mabar: Same as ParametersExtension check

src/DI/Compiler.php
```diff
@@ -192,8 +200,14 @@ public function addExportedTag(string $tag)
...
-		if (isset($this->extensions[self::DI])) {
-			$this->extensions[self::DI]->exportedTypes[$type] = true;
+		$extension = $this->extensions[self::DI];
+
+		if ($extension instanceof Nette\DI\Extensions\DIExtension) {
+			$extension->exportedTypes[$type] = true;
+		} else {
+			throw new InvalidConfigurationException(
+				"Parameter 'exportedTypes' in extension '" . self::DI . "' does not exist."
+			);
```
mabar: Same as ParametersExtension check

janbarasek: Fixed. Thanks.

src/DI/Compiler.php
janbarasek: Fixed. Thanks.

dg:
Thanks

PR #224, opened 2019-11-22
Allow fix of nette/php-generator

maxhelias:
Just allow minor constraint 😄

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
 		"php": ">=7.1",
 		"ext-tokenizer": "*",
 		"nette/neon": "^3.0",
-		"nette/php-generator": "^3.2.2",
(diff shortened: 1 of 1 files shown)


maxhelias:
The failure of travis is not related

dg:
Why?

maxhelias:
Version 3.3.1 is available and compatible, it would be cool to have it

dg:
See this https://getcomposer.org/doc/articles/versions.md#caret-version-range-

maxhelias:
And the failure was due to the 3.2.0 which has the class `PhpLiteral` in final and which does not have it anymore in 3.2.2

maxhelias:
Oh! Yes, sorry for the inconvenience

dg:
:-)

PR #225, opened 2019-12-08
Compiler: Self return type added.

janbarasek:
- new feature
- BC break? no

Added `self` to return.


dg:
I will add this when the minimum required version of PHP will be 7.4 due to return type covariance.

PR #233, opened 2020-02-26
Resolver: better exception message

janbarasek:
- bug fix
- BC break? no

In case of neon file:

```yaml
services:
   tagManagerAccessor:
      factory: Namespace\TagManagerAccessor
```

Nette throws user unfriendly exception:

`Class 'Namespace\TagManagerAccessor' not found.`

I think if the class does not exist, but the interface does, Nette should throw more user-friendly exception like:

`Service 'tagManagerAccessor': Class 'Namespace\TagManagerAccessor' not found. Interface 'Namespace\TagManagerAccessor' can not be used in 'factory' section as service.`

Thanks.

### DIFF
--- a/src/DI/Resolver.php
+++ b/src/DI/Resolver.php
@@ -134,10 +134,10 @@ class Resolver
 
 		} elseif (is_string($entity)) { // class
 			if (!class_exists($entity)) {
-				throw new ServiceCreationException(
-					'Class ' . $entity . ' not found.'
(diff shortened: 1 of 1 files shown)


dg:
I don't like part `Class 'Namespace\TagManagerAccessor' not found.`. What about `Interface 'Namespace\TagManagerAccessor' can not be used as 'factory', did you mean 'implement'?`

janbarasek:
@dg Your message makes sense. Implemented.

dg:
Thanks

PR #236, opened 2020-04-18
Container phpstan generics

mabar:
- new feature
- BC break? no
- doc PR: not needed

It helps phpstan understand that returned instance is always instance of $class

### DIFF
--- a/src/DI/Container.php
+++ b/src/DI/Container.php
@@ -265,6 +265,9 @@ class Container
 	 * Creates new instance using autowiring.
 	 * @return object
 	 * @throws Nette\InvalidArgumentException
+	 * @template T of object
+	 * @phpstan-param class-string<T> $class
+	 * @phpstan-return T
 	 */
 	public function createInstance(string $class, array $args = [])
 	{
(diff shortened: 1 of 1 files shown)


dg:
This would be especially useful for getByType(), wouldn't it?

mabar:
You are right, added

dg:
Hey, isn't this exactly what phpstan does? https://github.com/phpstan/phpstan-nette/blob/master/src/Type/Nette/ServiceLocatorDynamicReturnTypeExtension.php

mabar:
Yes, but native support is imho better than plugin. This way also other static analysis tools and IDEs can understand it. 
Phpstan and psalm are able to share their prefixed docblocks (e.g. psalm uses `@phpstan-param`
IDEs could possibly have native generics support in future, so .phpstorm.meta.php will be not needed (there is already phpstorm plugin https://github.com/Haehnchen/idea-php-generics-plugin)

dg:
I agree that native support is better, but this little bit premature - I'd rather wait until there is an official annotation. Because I think that everybody who analyzes projects in Nette has this plugin.

mabar:
ok

hrach:
Note: these type of annotations also helps devs to easily understand what the method will do.

dg:
@hrach how?

hrach:
It's easier to read this generic type than undestand it from description (because returning an object tells me nothing).

dg:
You're probably an exception, I think the vast majority of PHP programmers don't understand generics yet.

PR #237, opened 2020-04-27
SearchExtension: Exception when class is not autoloaded.

janbarasek:
- bug fix
- BC break? no

Throw exception when class is not autoloaded.

For example I set find directory to project root, but Robot loaded is mapping `app` dir only.


dg:
Thanks

PR #242, opened 2020-06-21
Added support for alternative annotation syntax

peldax:
- new feature
- BC break? most likely not

Supports `@param array<Abc\Xyz> $param` as alternative to `@param Abc\Xyz[] $param`. Code is duplicated, you might want to adjust the conditions a little.


dg:
Implemented, #277

PR #246, opened 2020-10-10
Resolver: More better exception in case of accessor or factory is defined as regular service.

janbarasek:
- bug fix
- BC break? no

In the case of the interface (factory or accessor) is defined as regular service, DI should throw a more helpful exception.

For example (broken code in extension):

```php
$builder->addDefinition($this->prefix('doctrineHelperAccessor'))
	->setFactory(DoctrineHelperAccessor::class);
```

This code should be rewritten to:

```php
$builder->addAccessorDefinition($this->prefix('doctrineHelperAccessor'))
	->setImplement(DoctrineHelperAccessor::class);
```

Thanks!


dg:
I'm closing, the issue is already outdated.

Issue #229, opened 2020-01-13
"Prevent merging" conflicts with Schema validation


dakujem:
Version:
+ nette/di: master|3.0.2
+ nette/schema: 1.0.2
+ nette/application: 3.0.3

**The exception**
> Nette\DI\InvalidConfigurationException\
> The option 'application › mapping › _prevent_merging' expects to be string or array, bool given.

![image](https://user-images.githubusercontent.com/443067/72248944-a420dc80-35f8-11ea-88a7-651eef088166.png)

I believe the Schema is trying to validate `_prevent_merging => true` as one of the `mapping` options, which should not be the case.

**The configuration**
```neon
application:
    errorPresenter: Error
    mapping!:
        *: App\*Module\Presenters\*Presenter
```

Gappa:
I've encoutered the same problem with scanDirs, which can be currently disabled, added to, but not overwritten:

``` yaml
application:
    scanDirs!:
        - myCustomDir/
```

Which results to:

```
Nette\DI\InvalidConfigurationException
The option 'application › scanDirs › _prevent_merging' expects to be string, bool given.
```

Relevant packages:

```
nette/application     v3.0.3 
nette/di              v3.0.3
nette/schema          v1.0.2
```

Gappa:
I just found out that (at least for scanDirs) the behaviour is different from user defined directives - by specifing a value the default one gets overridden:

So this:

``` yaml
application:
    scanDirs:
        - myCustomDir/
``` 

actually results to `['myCustomDir/']` and **not** `['%appDir%', 'myCustomDir/']` as [I was expecting](https://doc.nette.org/en/3.0/configuring#toc-automatic-registration-of-presenters) and trying to override.

So, is this a bug afterall or just a change in behaviour?

dg:
@Gappa it seems like a bug

f3l1x:
I have just hit the same error. Also with application.scanDirs and application.mapping.

dg:
fixed

f3l1x:
❤️

PR #250, opened 2020-12-01
Prevent merging

f3l1x:
- issue: #229
- BC break: nope
- doc PR: nope

----

This PR provides failing tests for #229.


dg:
thanks

PR #257, opened 2021-02-18
Fix silencing file_get_contents file not not found warning

b1rdex:
- bug fix   
- BC break? no

Hi. `file_get_contents` might produce a warning if the file isn't found. There is a comment `@ - file may not exist`, but `@` was mistakenly placed on `unserialize`.


JanTvrdik:
Does't silencing the `unserialize` also silence the `file_get_contents` call?

b1rdex:
Wow. You're right. https://3v4l.org/NRLMV

PR #258, opened 2021-03-16
SchemaExtension

mabar:
- new feature 🎉 
- BC break? no
- doc PR: Okay, but later please.

Any key in neon can be validated with neon-defined [nette/schema](https://github.com/nette/schema), via `SchemaExtension`:

```neon
extensions:
	schema: Nette\DI\Extensions\SchemaExtension

schema:
	string: string()
	structure: structure([
		string: string(),
		stringWithDefault: schema(string("default value"), required(false))
		int: int(),
		float: float(),
		bool: bool(),
		array: arrayOf(string())
		list: listOf(string())
		type: type("string|int")
		schema1: schema(string())
		schema2: schema(string(), nullable())
		schema3: schema(string(), nullable(), required(false))
		schema4: schema(int(), min(10), max(20))
```

Behavior differs from direct use of nette/schema
- all keys defined in schema are required by default (should make the most common usecase of neon-defined schema simpler)
- additional configuration which is done via chained method can be achieved via `schema()`, eg. `Expect::int()->min(10)->max(20)` -> `schema(int(), min(10), max(20))`

Requirements of schema above could be satisfied with this example configuration:

```neon
string: string
structure:
	string: text
	int: 123
	float: 123.456
	bool: true
	array: [key: string, anotherString]
	list: [string, anotherString]
	type: string
	schema1: string
	schema2: null
	#schema3 is not required
	schema4: 15
```

Keys not defined in schema are allowed and accept any value just like before this PR. So e.g. this is allowed:
```neon
services:
	- Example
```

TODOs
- docs
- bootstrap integration

Big thanks to author of original extension which validates parameters in PHPStan, @ondrejmirtes

### DIFF
--- a/src/DI/Extensions/SchemaExtension.php
+++ b/src/DI/Extensions/SchemaExtension.php
@@ -1,0 +1,142 @@
+<?php
+declare(strict_types=1);
+
+namespace Nette\DI\Extensions;
+
+use Nette;
+use Nette\DI\Definitions\Statement;
+use Nette\Schema;
+use Nette\Schema\Expect;
+
+/**
+ * Based on ParametersSchemaExtension from PHPStan
+ *     https://github.com/phpstan/phpstan-src/blob/6d523028e399c15dc77aec3affd2ea97ff735925/src/DependencyInjection/ParametersSchemaExtension.php
+ * @property-read Statement[] $config
+ */
+final class SchemaExtension extends Nette\DI\CompilerExtension
+{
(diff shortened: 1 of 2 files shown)


mabar:
Build failures seem unrelated to the PR.

I also had to remove deprecation from Compiler::getConfig() as it seems to be only way how can extension access whole config.

mabar:
@dg Are any changes needed or are you just waiting for next release or smth?

PR #261, opened 2021-09-11
Add typehints to Container::getByType objects (master)

Rixafy:
- new feature
- BC break? no


janbarasek:
Can it be reverted? Rel #263

dg:
@janbarasek this PR is related only to Container::getByType()

PR #262, opened 2021-09-24
typo

dakur:
- bug fix
- BC break? no


dakur:
I was wondering why all tests were failing... 🙂

Issue #199, opened 2019-05-10
More self-explaining message when constructor and factory parameters are mismatching


jkuchar:
- **Is your feature request related to a problem?**  
I would like to propose **more self-explanatory exception message** for case when constructor and magic factory implementation does not match.
- **Explain your intentions.**  
I found that it is not always clear that constructor and magic fatory are **bind together using parameter names**, not types. Which is logical from implementation perspective (no ambiguity), however counter-intuitive from user perspective.

originally implemented in #99

# Proposal

Instead of saying just 

![image](https://user-images.githubusercontent.com/133822/57511952-b11f5a80-730a-11e9-967b-cb237b97fd54.png)

I would say `Cannot implement ..\Components\HeaderFactory::create(), because parameters ($breadcrumbs, $logout) of factory method are not matching constructor of ...\Components\Breadrumbs ($breadcrumbsComponent, $logoutComponent).`

jkuchar:
It you agree with this proposal I will send PR for this.

dg:
👍

jkuchar:
Approved by @dg. @dakur Could you please make look at this? :)

dakur:
Here we go.. #284

PR #284, opened 2023-01-17
More self explanatory message for factory and service mismatch

dakur:
- bug fix / new feature?  #199 
- BC break? no
- doc PR: N/A


dg:
Thanks!

PR #285, opened 2023-02-20
fix Container::getByType() compatibility with PHPStan

JanTvrdik:
Without specifying the upper bound, PHPStan fails to resolve the return type.


dg:
Thanks

ondrejmirtes:
Generics should be verified with files like these analysed by PHPStan :) https://github.com/laravel/framework/blob/10.x/types/Support/Collection.php

PR #293, opened 2023-09-09
Support `list<Type>` syntax for autowiring a collection of services

spaze:
- bug fix / new feature?

New feature
[Autowiring a collection of services](https://doc.nette.org/en/dependency-injection/autowiring#toc-collection-of-services) already supports `@param Type[] $foo` and `@param array<int, Type>`, and using `@param list<Type> $foo` is basically the same thing in this case. For some people, like me, using `list<>` better indicates that the key of the array can be ignored (or can be assumed sequential).

- BC break? yes/no

no

- doc PR: nette/docs#???  

https://github.com/nette/docs/pull/1031


dg:
thanks

PR #294, opened 2023-09-14
ServicesExtension: locator support Statement

h4kuna:
- bug fix / new feature?   new feature
- BC break? no
- doc PR: nette/docs#???  

## Example

In example I used classes from my library, if you want to try, let's use `composer require h4kuna/number-format:dev-mater`, but i think it is not necessary.

I want to define custom collection of formats

```neon 
services:
	format.date:
		factory: h4kuna\Format\Date\Formatters\DateTimeFormatter('j.n.Y')
		autowired: false
	format.time:
		factory: h4kuna\Format\Date\Formatters\DateTimeFormatter('H:i:s')
		autowired: false

	date.formats: h4kuna\Format\Date\FormatsAccessor(
		date: @format.date
		time: @format.time
	)

	# set to template, but it is not need for example
	latte.latteFactory:
		setup:
			- addFilter('date', @date.formats::get('date'))
			- addFilter('time', @date.formats::get('time'))
```

The original way is too chatty, the patch allow
```neon 
services:
	date.formats: h4kuna\Format\Date\FormatsAccessor(
		date: h4kuna\Format\Date\Formatters\DateTimeFormatter('j.n.Y')
		time: h4kuna\Format\Date\Formatters\DateTimeFormatter('H:i:s')
	)
```

I don't  need class with option `autowired: true` explicitly to register to container. But behavior is same like above.

What do you think?

### DIFF
--- a/src/DI/Extensions/ServicesExtension.php
+++ b/src/DI/Extensions/ServicesExtension.php
@@ -186,6 +186,16 @@ final class ServicesExtension extends Nette\DI\CompilerExtension
 		}
 
 		if (isset($config->references)) {
+			foreach ($config->references as $name => $reference) {
+				if ($reference instanceof Statement) {
+					$config->references[$name] = '@' . $this->getContainerBuilder()
+							->addDefinition(null)
+							->setFactory($reference)
+							->setAutowired(false)
+							->getName();
+				}
+			}
+
 			$definition->setReferences($config->references);
 		}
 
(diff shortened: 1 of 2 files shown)


dg:
I have for a long time in the TODO to simplify the multi-factory. To make it defined the same way as a classic generated factory. Maybe like this:

```neon 
services:
	date.formats: 
		implements: h4kuna\Format\Date\FormatsAccessor
		contains:
			date: h4kuna\Format\Date\Formatters\DateTimeFormatter('j.n.Y')
			time: h4kuna\Format\Date\Formatters\DateTimeFormatter('H:i:s')
```

Compared to the syntax with parentheses, the advantage is that you can easily add, for example, `setup`:

```neon 
		contains:
			date: h4kuna\Format\Date\Formatters\DateTimeFormatter('j.n.Y')
			time: 
				factory: h4kuna\Format\Date\Formatters\DateTimeFormatter('H:i:s')
				setup:
					- foo
```

PR #308, opened 2024-02-09
Test for "Nette DI: suspicious dumping of objects DateTimeImmutable when generating the container"

kukulich:
`DateTimeImmutable` throws user notice since version 3.2.0:

```
E_USER_NOTICE: Nette DI: suspicious dumping of objects DateTimeImmutable when generating the container
```

PR contains failing test.


dg:
Thanks

PR #320, opened 2025-04-11
Enable to suppress Suspicious dumping error

survik1:
- bug fix / new feature?   new feature
- BC break? no

Nette/Di introduced warning about suspicious dumping of objects. While i understand the intention, this little warning forces as to rewritte significant amount of value objects that we just pass around. These value objects commonly implements specific interfaces, following pattern of private variable and getters. Even though we could spend time to rewrite everything into PHP 8.4 pattern of `public protected(set) $variable` properties (and hopefully don't run into any more complicated obstacle), I belive that this dev-oriented warning should be possible to suspend by dev.

I propose to add a class attribute that would prevent raising up this warning.

I was unable to add any tests as I don't know how to test trigger_error() function, but I am willing to add them if you point me in some direction.


survik1:
@dg Hello, could you please comment this pr in any way?

survik1:
@dg Hi, sorry for delay, I wanted to trim our code a bit.

We have Extension around Kafka. It parses configuration definitions from rather large neon file, completes it into KafkaConfig class and pass this class to extension services.

```php
class KafkaExtension extends CompilerExtension
{
    public function __construct()
    {
        $this->kafkaConfig = new KafkaConfig();
    }

    protected function loadConfiguration(): void
    {
        // ...
        foreach ($clusters as $clusterName => $clusterConf) {
            // ...
            $clusterTemplate = new ClusterTemplate($clusterName, [], $options, $callbacks);
            $this->kafkaConfig->addClusterTemplate($clusterTemplate);
            // ...
     }

     public function setupDefinitions(): void
     {
        $builder = $this->getContainerBuilder();
        $builder->addDefinition($this->prefix('producerFactory'))
            ->setFactory(ProducerFactory::class)
            ->setAutowired($this->config[self::AUTOWIRED] ?? true)
            ->setArguments([$this->kafkaConfig, $this->configurationHelper]);
         // ...
     }
}
```

KafkaConfig class is not a service, its more of ValueObject for all configuration data. 
```php
final readonly class ConsumerFactory
{
    public function __construct(
        private KafkaConfig $kafkaConfig, 
        private ConfigurationHelper $configurationHelper
    ) {
    }

    public function createConsumer(string $consumerName, ?Callbacks $callbacks = null, array $customConf = []): KafkaConsumer
    {
        $consumerTemplate = $this->kafkaConfig->getConsumerTemplate($consumerName);
        // ...
        return new KafkaConsumer($conf, $consumerTemplate);
    }
}
```

So yeah, its for ValueObjects.

And its kinda hard to modify it because I don't even understand your intentions here.

survik1:
@dg Hello, could you please help us? We would like to upgrade to PHP 8.4, but we are unable to deal with this notice.

survik1:
Solved with help of AI so I dont really care anymore

PR #328, opened 2025-12-22
Use Tracy `Bluescreen::$keysToHide` in the DI Tracy panel

### LINKED ISSUE #327: The Tracy DIC panel doesn't respect `Debugger::keysToHide`
(... which may be by design, but hear me out 😇)

Version: 3.2.5

### Bug Description
I've seen quite a few production apps with the Tracy bar enabled. That itself is quite bad, but if there's a database connection object displayed in the DIC panel in the Tracy bar, the situation can easily go from bad to worse, because the object has a database connection password. And while database servers usually do not accept connections from the Internet, sometimes there's phpMyAdmin or similar, especially on shared hosting servers, making the password suddenly useful again.

So I'm trying to see if Tracy or the nette/di could somehow help the developers who accidentally left Tracy enabled, as sort of an additional defense layer. I'm not sure if this is a bug report or a feature request, or even a discussion, so sorry if I filed it wrong.

I know there are several places where you can configure `keysToHide`, like the [`Debugger`](https://github.com/nette/tracy/blob/7543389d84c55e73b4fd2c4c9412da3aaeda3b18/src/Tracy/Debugger/Debugger.php#L93) class, the [`BlueScreen`](https://github.com/nette/tracy/blob/7543389d84c55e73b4fd2c4c9412da3aaeda3b18/src/Tracy/BlueScreen/BlueScreen.php#L37-L40) class in Tracy, and there's also a [Nette configuration](https://tracy.nette.org/cs/configuring#toc-nette-framework) as described in the docs (which seems to be not working as expected https://github.com/nette/tracy/issues/608). I have expected that I could somehow specify `keysToHide` somewhere (possibly in one place, maybe the Nette config?) and the password would be hidden in the DIC panel as well, provided that the `password` key would be there.

But the panel is a completely separate thing which uses Tracy's `Dumper::toHtml` in a [completely independent way](https://github.com/nette/di/blob/51b3399f12bdf803d40964859453db1bdcaaffd1/src/Bridges/DITracy/dist/panel.phtml#L61):
```php
<?= Dumper::toHtml($instances[$name], [Dumper::COLLAPSE => true, Dumper::LIVE => true, Dumper::DEPTH => 5]) ?>
```

As a proof of concept, I have added
```php
Dumper::KEYS_TO_HIDE => ['password']
```
to the line above, and the password was hidden. But that doesn't seem like an elegant way at all.

I'm open to ideas and can then try to prepare a pull request if needed.

### Steps To Reproduce
Have a Nette app that uses a database connection, enable Tracy, check the DIC panel and find the connection object. It will display a database password, even if you configure Nette to hide the `password` key.

### Expected Behavior
It would be cool if the panel also respected `keysToHide` configured in Nette config.

### Possible Solution
- #328
- Relevant Tracy PR https://github.com/nette/tracy/issues/609, sets the new `keysToHide` array when `tracy.keysToHide` config is set

spaze:
- bug fix / new feature? new
- BC break? no
- doc PR: will prepare if this will be merged

Relevant to https://github.com/nette/tracy/pull/609

Close #327


spaze:
Hey David @dg, sorry to be *that guy* 😅 but did you have a chance to check this, is there anything that needs tuning or something? Thank you!

